PR #18, opened 2023-12-11
FR-5869 -  Improve test coverage

upils:


sil2100 (approved):
This took a while to review, but I'm basically happy with the contents. There were a few smaller things I saw but nothing actually needing changes.

...and thanks for noticing the missing `services/` directory which I clearly forgot to `git add`. Apologies for that.

I think we need to get better with leaving meaningful descriptions on pull requests though. Since we might want to know in the future what a given PR merge was about. But that's a thing for the future.

PR #22, opened 2024-03-11
Use new ubuntu-image feature to build a rootfs easily

upils:
Ubuntu-image now enable building a rootfs (only) without using the `--thru` flag to stop the build. Moreover, the `artifacts` key is not mandatory anymore and can be omitted.

Imagecraft can now call ubuntu-image in a cleaner way.

This PR also fixes LP#2045336

Fixes: FR-6416


sil2100 (approved):
This is good! Quick and easy.

PR #23, opened 2024-03-19
Use the package-repositories stanza in imagecraft.yaml

upils:
Use the package-repositories stanza, relying on craft-archives and adapting the various classes to fit our needs.

This PR also contains:
- several improvements on the Project model
- a greater test coverage
- code cleaning and refactoring

Fixes: FR-5873


upils:
@cmatsuoka @tigarmo If you are bored or would like to sit back and relax, feel free to take a look at this PR :D. For now the implementation is done without any changes to craft-archives.

sil2100 (approved):
LGTM! I'm happy with how this turned out. I am still thinking about image_definition.py being a model, but it is a bit different than the craft model idea. Anyway, let's merge as is!

PR #27, opened 2024-03-28
Sync imagecraft with current imagecraft.yaml proposition

upils:
Get the current proposition of the YAML configuration file supported in the current imagecraft PoC. Specific fields needing rework in other libs (ex. `package-repositories`) are excluded.

Fixes: FR-7199

### DIFF
--- a/imagecraft/ubuntu_image.py
+++ b/imagecraft/ubuntu_image.py
@@ -18,7 +18,7 @@
 
 import subprocess
 
-import yaml
+from craft_application.util import dump_yaml
 from craft_cli import emit
 from pydantic import BaseModel, Field
 
@@ -158,13 +158,12 @@ def generate_image_def_yaml(  # noqa: PLR0913
             extra_packages=extra_packages_obj,
         )
 
-    return yaml.dump(
+    return dump_yaml(
         image_definition.dict(
             exclude_unset=True,
             exclude_none=True,
             by_alias=True,
         ),
-        sort_keys=False,
     )
 
 
--- a/imagecraft/ubuntu_image.py
+++ b/imagecraft/ubuntu_image.py
@@ -18,9 +18,96 @@
 
 import subprocess
 
+import yaml
 from craft_cli import emit
+from pydantic import BaseModel, Field
 
 from imagecraft.errors import UbuntuImageError
+
+
+def _alias_generator(s: str) -> str:
+    return s.replace("_", "-")
+
+
+class Snap(BaseModel):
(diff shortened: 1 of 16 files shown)


imagecraft/ubuntu_image.py
```diff
@@ -18,13 +18,101 @@
...
+    class Config:
+        """Pydantic model configuration."""
+
+        validate_assignment = True
+        allow_mutation = True
+        allow_population_by_field_name = True
+        alias_generator = _alias_generator
+
+
+class ImageDefinition(BaseModel):
```
sil2100, line 91: I like this idea! I'm pretty sure we discussed that at some point, but have you looked into if it would be possible to somehow have some common, 'shared' schema for both ubuntu-image and imagecraft? You know, to have a single source of truth for the image definition schema? Sounds like non-trivial, but maybe..? This is good so far though!

imagecraft/ubuntu_image.py
```diff
@@ -108,7 +209,7 @@ def ubuntu_image_pack(
 ) -> None:
     """Pack the primed image contents into an image file."""
     cmd: list[str] = [
-        "./ubuntu-image",
+        "ubuntu-image",
```
sil2100, line 212: An idea for the future: since we're now depending on ubuntu-image to be installed, could we maybe add a check to see if ubuntu-image is installed and available at the start of imagecraft?

sil2100 (approved):
Okay, this looks good! Let me merge it :) Left some inline comments but those are mostly just discussion topics, nothing actionable.

imagecraft/ubuntu_image.py
upils: Oh yeah that is the first idea I had! :D I did this kind of things at my previous job, using protobuf. So I had a single set of protobuf files, and used the generator to derive the model in different languages. This is very handy to make sure models are kept up to date everywhere but it comes with a non-negligible overhead and complexity. I also thought about looking into tools that could convert a go struct to a pydantic class. This is more specify but could be easier to use considering our use case. 

Our image definition model is not evolving a lot, so I am not sure this is worth the investment though. I will keep it on my todo list for now and see if it becomes increasingly needed.

imagecraft/ubuntu_image.py
upils: I need to learn a bit more about snap dependencies. Because I declared ubuntu-image as a staged-snap so I suppose it is embedded in the imagecraft snap, so I guess it is necessarily present, right?

PR #30, opened 2024-04-11
Package imagecraft as a snap

upils:
Package imagecraft as a snap to make it easily usable and reach parity with ubuntu-image

Fixes: FR-7305


upils:
@cmatsuoka could you review please? It looks like I cannot add you as the reviewer :thinking:. Maybe because you do not have maintainers permissions on the project.

upils:
The linter issue is fixed in another branch, it can be ignored.

setup.cfg
```diff
@@ -0,0 +1,76 @@
+[metadata]
+name = imagecraft
+version = attr: imagecraft.__version__
+description="Create Ubuntu bootable images."
+url = https://github.com/canonical/imagecraft
+project_urls =
+    Source = https://github.com/canonical/imagecraft.git
+    Issues = https://github.com/canonical/imagecraft/issues
+author = Canonical Ltd.
+author_email = snapcraft@lists.snapcraft.io
```
sil2100, line 10: Should we maybe put the ubuntu-devel mailing list here instead or something..?

setup.cfg
```diff
@@ -0,0 +1,76 @@
+[metadata]
+name = imagecraft
+version = attr: imagecraft.__version__
```
sil2100, line 3: My python-brain does not comprehend this version number handling..!

snap/snapcraft.yaml
```diff
@@ -0,0 +1,68 @@
...
+    instead.
+adopt-info: imagecraft
+confinement: classic
+grade: devel
+license: GPL-3.0
+source-code: https://github.com/canonical/imagecraft
+
+apps:
+  imagecraft:
+    command: bin/python $SNAP/bin/imagecraft
```
sil2100, line 16: ...should we just use `bin/python3` instead, explicitly?

sil2100 (approved):
This is looking good. Small thing to fix either before merging or later (see inline), but otherwise snap is building and working as expected. Thanks!

snap/snapcraft.yaml
upils: I think we can yes. This is another thing I left "as is" because it is done this way in every but one craft apps I looked at and was not sure how much of a convention this is. On this one I think you are right. I will fix it.

setup.cfg
upils: I discovered this is the idiomatic way to use the `__version__` attribute defined/computed in the package. See https://packaging.python.org/en/latest/guides/single-sourcing-package-version/ (specifically the `Note`).

setup.cfg
upils: I thought about that too but I used the same email I found in other craft apps. I suppose we probably won't package and distribute the tool under another format than a snap, so I am not even sure this value will ever be used by a user to contact us. I see this as part of the "template" of a `setup.cfg` file to use when packaging a craft snap. 

@cmatsuoka what do you think about it?

sil2100 (approved):
LGTM!

PR #47, opened 2024-07-08
Migrate to core24 and python 3.12

upils:
Migrate imagecraft to core24 to avoid relying on 2 cores (since ubuntu-image is also now using core24) and so use python3.12 in the snap.

Python3.12 is explicitly used for now because otherwise python3.10 is picked up during the build. This will need to be removed at some point.

Also add a first very basic spread test to:
- make sure we test the snap is building and running fine
- prepare for more tests


upils:
Even though every CI check is green, the build of the snap is broken. This needs to be fixed before merging.

upils:
This is blocked by https://github.com/canonical/craft-parts/issues/786

upils:
@sil2100 2 tests are still failing because we do not have the runners setup yet. Sergio from the snapd team should set them up today or this week.

Makefile
```diff
@@ -13,4 +13,4 @@ freeze-requirements:  ## Re-freeze requirements.
 
 .PHONY: snap
 snap:  ## Create a Imagecraft snap
-	@snapcraft clean && snapcraft --use-lxd
+	@snapcraft clean --use-lxd && snapcraft --use-lxd
```
sil2100, line 16: I didn't even know that the clean state also required passing the provider to use...

pyproject.toml
```diff
@@ -12,9 +12,9 @@ classifiers = [
     "Development Status :: 1 - Planning",
     "License :: OSI Approved :: GNU General Public License (GPL)",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.12",
 ]
-requires-python = ">=3.10"
+requires-python = ">=3.11"
```
sil2100, line 17: Are we sure 3.11 is the minimal, or should that be 3.12 here as well?

setup.cfg
```diff
@@ -9,17 +9,17 @@ project_urls =
...
     Intended Audience :: Developers
     License :: OSI Approved :: GNU General Public License v3 (GPLv3)
     Operating System :: POSIX :: Linux
     Programming Language :: Python :: 3
-    Programming Language :: Python :: 3.10
+    Programming Language :: Python :: 3.12
 
 [options]
-python_requires = >= 3.10
+python_requires = >= 3.11
```
sil2100, line 22: Same question as above.

snap/snapcraft.yaml
```diff
@@ -11,42 +11,52 @@ grade: devel
...
+  - python3.12-minimal
 
 parts:
   imagecraft-libs:
     plugin: nil
     build-attributes:
       - enable-patchelf
     stage-snaps:
-      - ubuntu-image
+      - ubuntu-image/latest/edge
```
sil2100, line 45: I know we're still in testing, but is using edge here always a good idea? What's the concept here? Is it because the stable ubuntu-image isn't core24-base yet, so temporary?

sil2100 (dismissed):
LGTM! Do we want to wait with merging it until Sergio is done? Also, left some inline comments, but all might be irrelevant (just questions).

Makefile
upils: I didn't either until someone from the craft team asked if I did that. I have to say I did not checked if this is really making a difference but at least we are "sure" this is cleaning what we are using to build.

pyproject.toml
upils: I was also wondering. Because if we consider the tool is to be used only as a snap then we do not really care for previous python versions. I am personally in favor of dropping support for python 3.11 (less linter fights, less tests to run, etc.) but I may miss some contraints. What do you think about that?

snap/snapcraft.yaml
upils: Sorry I will add a comment. This is "needed" because otherwise installing the current stable version of ubuntu-image pulls core22 and thus python3.10. And according to the craft team this might be the cause of the issue I had (python3.10 was picked up when building the `imagecraft` part instead of the theoretically default python3.12). I am really not sure about that so as soon as we land a core24 based ubuntu-image in latest/stable I will remove that because I do not want imagecraft to rely on the edge channel of ubuntu-image.

upils:
> LGTM! Do we want to wait with merging it until Sergio is done? Also, left some inline comments, but all might be irrelevant (just questions).

Yes I would like to wait for the setup of the runners to make sure this is working fine. I copied the conf from ubuntu-image but I may have messed it up so I would like to be sure the whole thing will properly work before merging.

upils:
@sil2100 I have added snapd-testing-tools as a subtree (like in ubuntu-image) so do not worry about the files added in `tests/lib/external/`. As previously mentioned, due to https://github.com/canonical/craft-parts/issues/786, we cannot really merge this sadly because otherwise the tool will only work on 24.04. 

With Sergio we are still trying to understand why one of the runner is not working properly but otherwise once we decide if we keep python3.11 or not, the code looks ready for a last review.

pyproject.toml
sil2100: I'd be +1 on that. Doesn't have to be changed as part of this PR though.

sil2100 (approved):
I don't see anything controversial in these changes, LGTM per our discussions. Of course there's a lot, but it's all due to the snapd-testing-tools, as mentioned.

PR #50, opened 2024-07-25
Document commands

upils:
Document imagecraft's available commands.

We are dynamically "calling" imagecraft to extract available commands and try to make sure we will keep up with new commands from the craft framework automatically.


docs/reference/commands/pack.rst
```diff
@@ -0,0 +1,48 @@
+
+.. _ref_commands_pack:
+
+pack
+====
+
+Process parts and create the final artifact.
```
sil2100, line 7: Should we maybe be a bit more specific about what this means? Since right now it's rather quite generic. I think it would be best if we at least said that at this stage, in most cases, this is when the 'image' in image-craft is created - usually a disk with partitions. Of course in later imagecraft versions we might support other formats, like tarballs or something else crazy.

Was thinking about maybe we should also mention that `ubuntu-image pack` is used for this stage? I know it's an implementation detail, but for people that are encountering errors that might be a useful fact. I'm not very strongly opinionated about this though.

docs/reference/commands/stage.rst
```diff
@@ -0,0 +1,52 @@
+
+.. _ref_commands_stage:
+
+stage
```
sil2100, line 4: I'm wondering what this means to us basically. Not much used for the imagecraft lifecycle.

docs/reference/commands/build.rst
```diff
@@ -0,0 +1,51 @@
+
+.. _ref_commands_build:
+
+build
+=====
+
+Build artifacts defined for a part. If part names are specified only
+those parts will be built, otherwise all parts will be built.
```
sil2100, line 8: I'm thinking if we maybe should mention a bit about how the parts for an imagecraft image look like here. Maybe we should think about it and return to it later though.

.github/workflows/snapcraft_yaml.yml
```diff
@@ -1,25 +0,0 @@
-name: Has snapcraft.yaml been updated?
```
sil2100, line 1: This has been recently added (with the core24 PR). I suppose the removal is intentional? I have no strong opinions. I mean, such a check is nice and we could consider it non-blocking until we go a bit more official.

sil2100 (approved):
Okay, left some inline comments, but they're more like questions and propositions. The general gist of the documentation is looking good.

.github/workflows/snapcraft_yaml.yml
upils: It is intentional. I gave some details in the commit removing it. We now dynamically set the version on the snap at build time. So relying on changes on the `snapcraft.yaml` to make sure we incremented the version is not making sense anymore. I need to talk to other teams working on craft apps to understand how they manage to set a specific version/tag and then promote it. This is not our priority for this cycle but will need to be addressed before the first beta or candidate version of imagecraft.

docs/reference/commands/pack.rst
upils: Every file in `docs/reference/commands/` is generated. Since we do not define any specific command, all of this comes from some craft libs. However your suggestions should probably go into the `explanation` section or be part of the tutorial.

docs/reference/commands/stage.rst
upils: Currently we only do any real work at the build and pack stages. But since this is a craft apps, these commands are available anyway. Moreover, the craft framework could do some magic with some of these commands that users might be interested in, so I kept it. This is more about making sure the experience is consistent with other crafts than only documenting what is really useful to us.

docs/reference/commands/build.rst
upils: See previous comments. I agree with you explaining what is going on is valuable but it should probably be in the `explanation` section.

PR #51, opened 2024-07-30
Make the snap able to run on systems without python3.12

upils:


upils:
Some CI jobs failed, but because gnu.org is down. Will retry later but it should be good.

sil2100 (approved):
Can't wait for this to be fixed in craft libs.

PR #54, opened 2024-08-06
Add basic spread tests

upils:
Add a real spread test for on amd64 and one arm64 builds on 20.04, 22.04 and 24.04

Fixes: FR-6129


upils:
Tests are failing on 20.04 due to a "bug" in craft-parts lib (not aware it is running in a snap) but otherwise they are building fine.

tests/spread/amd64/classic/task.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+kill-timeout: 3h
+
+execute: |
+    mkdir -p /tmp/${SPREAD_VARIANT}/mount
+    pushd ${SPREAD_VARIANT}
+
+    imagecraft --verbosity debug pack --destructive-mode
+
+    IMG_NAME=pc.img
```
sil2100, line 11: I guess this is unused right now? Or did you mean on using this instead of `pc.img` directly?

sil2100 (approved):
All good. One inline comment but that's more of a cosmetic thing.

tests/spread/amd64/classic/task.yaml
upils: Correct, I meant on using this var. I will fix it. It will need to be reworked when we are able to better control the artifacts. Here I am lucky that we create a img called `pc.img` in both cases.

PR #55, opened 2024-08-07
Migrate to craft-application 3.X

upils:
Fixes: FR-8576


upils:
Hey @tigarmo @cmatsuoka would you mind also taking a look to make sure the usage of the craft libs looks sound? I am doing a ugly hack with the package-repositories field. I am not sure I can do much better/cleaner until we rework the way we specify the package repositories.

imagecraft/models/project.py
```diff
@@ -226,16 +187,55 @@ def _validate_all_platforms(cls, platforms: dict[str, Any]) -> dict[str, Platfor
 
         return platforms
 
+class Project(YamlModelMixin, BuildPlanner, BaseProject):
+    """Definition of imagecraft.yaml configuration."""
+
+    series: str
```
sil2100, line 193: I think this is good for now, but per our recent discussions we really need to see about dropping `series` and just building per `build_base`. Yes, this limits us to only building series that are equal to the build environment, but this is already a constraint in livecd-rootfs - and in theory in other crafts too.

sil2100 (approved):
Looking good, lots of code moving to the framework I see.

PR #59, opened 2024-09-20
Rename the UbuntuSeed plugin as UbuntuBootstrap

upils:
This should avoid some confusion because various users have a different understanding of "seeding".
UbuntuBootstrap also reflects better what the plugin is doing.

Since we are doing that early in the life of the project I am not sure we should keep compatibility with the old plugin name.

### DIFF
--- a/docs/howto/code/basic_imagecraft.yaml
+++ b/docs/howto/code/basic_imagecraft.yaml
@@ -20,9 +20,9 @@ parts:
     source: https://github.com/snapcore/pc-gadget.git
     source-branch: classic
   rootfs:
-    plugin: ubuntu-seed
-    ubuntu-seed-pocket: updates
-    ubuntu-seed-germinate:
+    plugin: ubuntu-bootstrap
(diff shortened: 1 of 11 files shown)


sil2100 (approved):
I think ubuntu-bootstrap is already a better name. Still sucks that this makes many of the ubuntu-bootstrap parameters to be long (like `ubuntu-bootstrap-extra-snaps:`), but I guess not much we can do with that.

upils:
> I think ubuntu-bootstrap is already a better name. Still sucks that this makes many of the ubuntu-bootstrap parameters to be long (like `ubuntu-bootstrap-extra-snaps:`), but I guess not much we can do with that.

A nicer way would be to use sub-keys. I already did that for the seed but I understand that is not "recommended" for plugins. I also thought of using a short version of `ubuntu-bootstrap` but everything I can think of is either existing (Ex. `u-boot`...) or not very nicer (ex. `ubunstrap`, `u-bootstrap`)

PR #60, opened 2024-09-24
Migrate to pydantic 2

upils:
Migrate to pydantic V2 and adapt to new craft libraries.

Fixes: FR-8917


upils:
@cmatsuoka @tigarmo A quick review of this from the craft team would be greatly appreciated if you can spare some time!

tigarmo (approved):
nice :)

PR #62, opened 2024-10-07
docs: Document how to rebuild the snap in latest/stable

upils:
This should be used when vulnerabilities are present in dependencies.


sil2100 (approved):
Same here: all good.

PR #65, opened 2024-10-17
Improve the docs homepage

rkratky:
- Navig. grid fix
- Added info to HP
- Spellcheck fix


upils (approved):
Thanks. LGTM!

I was not even aware the "tiles" on the homepage grid were supposed to be clickable.

PR #70, opened 2025-01-08
work/CRAFT-3862-Implement-the-initial-Imagecraft-spec

upils:

-----

Specification https://docs.google.com/document/d/1pTXN9LIFu4F0KmCT-iK2-_gGKELapKxvEj0-74xdr3g/edit?tab=t.0


common.mk
lengau, line 1: There's a new version that fixes a uv-related bug @tigarmo found. https://github.com/canonical/starbase/blob/main/common.mk

lengau (dismissed):

lengau (approved):

HACKING.rst
```diff
@@ -1,116 +1,363 @@
...
+Imagecraft projects follow the `OneFlow`_ git branching model.
 
-Other tools
-###########
 
-Some other tools we use for code quality include:
+Branch names
+============
+
+The branch name should be brief and less than 200 characters.
```
mattculler, line 53:
```suggestion
The branch name should be brief - less than 200 characters.
```

HACKING.rst
```diff
@@ -1,116 +1,363 @@
...
 
-- Black_ for code formatting
-- pytest_ for testing
-- ruff_ for linting (and some additional formatting)
+Branch names are formatted as ``<category>/<name>``.
```
mattculler, line 55:
```suggestion
Branch names are formatted as ``<category>/<name>``.  Categories are listed below.
```

HACKING.rst
```diff
@@ -1,116 +1,363 @@
...
-* ``tests``: The union of ``unit-tests`` and ``integration-tests``
+``hotfix/X.Y``
+##############
 
-For each of these, you can see which environments will be run with ``tox list``.
-For example::
+For hotfixes to an existing minor release.
 
-    tox list -m lint
+For example, hotfixes for Testcraft 2.1 would go to a branch called
```
mattculler, line 97:
```suggestion
For example, hotfixes for Imagecraft 2.1 would go to a branch called
```

HACKING.rst
```diff
@@ -1,116 +1,363 @@
...
-Spread tests rely on [snapd-testing-tools](https://github.com/snapcore/snapd-testing-tools). To update the subtree of this project, run::
+For commits that merge another branch into the current branch.  See the
+`chore <#chore>`_ section for information on merge commit headers.
+
+``release/X.Y.Z``
+#################
+
+For commits that prepare for a release. These commits should update the
+`changelog <#changelog>`_.
+
```
mattculler, line 118: Should there be a section here describing `feature/<description>` branches?

common.mk
```diff
@@ -0,0 +1,286 @@
+# Common items for all Imagecraft Makefiles. Should only be edited in the `starbase` repository:
```
mattculler, line 1: Does it make sense to even edit in the name "Imagecraft" here?  If this file shouldn't be edited except in starbase.

imagecraft/models/project.py
```diff
@@ -85,93 +87,26 @@ def _preprocess_all_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
...
-                    ),
-                )
-
-            platforms[platform_label] = platform
-
-        return platforms
-
 
-class Project(BuildPlanner, BaseProject):
+class Project(BuildPlanner, BaseProject):  # type: ignore[misc]
```
mattculler, line 91: Does mypy complain about multiple inheritance under the category "misc"?

mattculler, line 91: I understand how the BuildPlanner and Project will have some of the same fields, so I can see why you'd make the Project inherit from the BuildPlanner.  But, a Project is not a BuildPlanner, and I expect inherited classes to have an "is-a" relationship, so this solution feels wrong.

To avoid duplication, perhaps a solution in the same vein could be to make a third class?  But as I'm trying to come up with a name for this third class I can see the dilemma - these things derive from the project file, but they aren't ProjectFiles either.  I'm not seeing a good solution.

If you don't see a good solution here either then feel free to ignore, and I guess we can chalk this up to pydantic twisting the intent of inheritance.

pyproject.toml
```diff
@@ -16,29 +20,96 @@ classifiers = [
...
-    "types-requests",
-    "types-urllib3",
+apt-oracular = [
+    # 2.9 for Oracular+
+    "python-apt>=2.9.0;sys_platform=='linux'",
+]
+apt-plucky = [
+    # 2.9 for Oracular+
+    "python-apt>=2.9.0;sys_platform=='linux'",
+]
```
mattculler, line 49: I like this choice of multiple `apt-` dependency targets - I just tentatively stepped in a similar direction in `craft-parts` (`apt-manual` with the launchpad URL, for manual local installation).

mattculler (dismissed):
1. A link to the spec would be helpful
2. Love to see thousands of lines deleted!
3. It looks like this PR includes both a merge from starbase and spec implementation - in the future, splitting those two large changes into separate PRs would make for easier review.

But looks good!  None of my inline comments are blockers.

common.mk
upils: Good catch!

HACKING.rst
upils: I have no strong opinion on this suggestion and the next one but this is coming directly from starbase and ideally I would like to stay as close as possible to it to ease future "rebases".

HACKING.rst
upils: I have opened an issue https://github.com/canonical/starbase/issues/323 to do that in starbase.

imagecraft/models/project.py
upils: No, it complains of `Definition of "build_base" in base class "BuildPlanner" is incompatible with definition in base class "Project"` (and the same for "base"), but indeed under the `misc` category :/. I am not a fan of using misc either since I understand it may hide other linting errors but I did not find a more precise category.

upils: Good point. I am a bit confused regarding the way the BuildPlanner should be used. Let's discuss that in overflow.

upils:
> 1. A link to the spec would be helpful
> 2. Love to see thousands of lines deleted!
> 3. It looks like this PR includes both a merge from starbase and spec implementation - in the future, splitting those two large changes into separate PRs would make for easier review.
> 
> But looks good! None of my inline comments are blockers.

@mattculler I added the spec in the PR description. The PR is indeed doing both but the actual spec implementation is very very small here (this is just about setting constraints on base and build_base) so I was not sure it was worth splitting.

imagecraft/cli.py
```diff
@@ -16,23 +16,12 @@
...
-import logging
-
 from imagecraft.application import Imagecraft
-from imagecraft.plugins import setup_plugins
 from imagecraft.services.service_factory import ImagecraftServiceFactory
 
 
 def run() -> int:
     """Command-line interface entrypoint."""
-    # Register our own plugins
```
tigarmo, line 28: 👍🏾

imagecraft/models/project.py
```diff
@@ -64,10 +57,19 @@ def _validate_platform_set(self) -> Self:
         return self
 
 
+BaseT = Literal["bare"]
+BuildBaseT = typing.Annotated[
+    Literal["ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "devel"] | None,
```
tigarmo, line 62: do you really need to support focal? Is that defined somewhere?

imagecraft/services/lifecycle.py
```diff
@@ -19,14 +19,10 @@
 from typing import cast
 
 from craft_application import LifecycleService
-from craft_parts import Features
 from overrides import override  # type: ignore[reportUnknownVariableType]
 
 from imagecraft.models.project import Project
 
-# Enable the craft-parts features that we use
-Features(enable_overlay=True)
```
tigarmo, line 28: where is this being enabled now? e.g. https://github.com/canonical/rockcraft/blob/main/rockcraft/application.py#L82

imagecraft/services/pack.py
```diff
@@ -36,38 +35,23 @@ def __init__(
...
-        platform_output.mkdir(parents=True, exist_ok=True)
-
-        ubuntu_image_pack(rootfs_path, gadget_path, str(dest), workdir_path)
-
-        img_paths = list_image_paths(workdir_path)
-
-        shutil.rmtree(workdir_path)
 
-        return [pathlib.Path(dest) / img for img in img_paths]
+        return []
```
tigarmo, line 54: task-wise, what's the breakdown here? Is this coming in a future task?

tigarmo (commented):
Thanks! I mostly reviewed the contents for the `imagecraft` directory, let me know if there's anything important outside of it that I should look at (I assumed most of it is the standard starbase/starflow stuff)

imagecraft/services/lifecycle.py
upils: Nowhere :). I will enable this (or most probably the partition feature first) as a following step when implementing following specs (the original one is not mentioning it so I want to keep it clean for now).

imagecraft/services/pack.py
upils: Correct. I still wanted to keep the ImagecraftPackService around for now though since it may be a central part in imagecraft. I am not sure if this is noise or not though. What do you think?

upils:
> Thanks! I mostly reviewed the contents for the `imagecraft` directory, let me know if there's anything important outside of it that I should look at (I assumed most of it is the standard starbase/starflow stuff)

Thanks! I am also interested in a review of the GH workflow changes and the spread-related ones (I previously had tests on various systems for amd64/arm64 that I removed for now since the tool is not doing much).

imagecraft/services/pack.py
tigarmo: I think this is fine

tigarmo (approved):

HACKING.rst
mattculler: No problem

imagecraft/models/project.py
mattculler: Sounds good, maybe just drop a comment to that effect.

mattculler: Discussed today - I think we have a handle on the various solutions and their compromises, and I'm fine with whatever direction Paul chooses here.

mattculler (dismissed):
Looking good

medubelko (dismissed):
@upils I fixed the build and did some reorganisation of the docs so they align with current practices. Everything looks good from my end.

upils:
> @upils I fixed the build and did some reorganisation of the docs so they align with current practices. Everything looks good from my end.

Great, thanks!

imagecraft/models/project.py
upils: This is not needed anymore since Project is not inheriting from BuildPlanner anymore.

tigarmo (approved):

.readthedocs.yaml
lengau, line 1: I can't see why the RTD builds are failing, but they're currently failing.

HACKING.rst
lengau, line 1: I know this is different from what we have in starbase, but I think @medubelko wants us to move over to `CONTRIBUTING.rst` so maybe we can do that here?

lengau (approved):

HACKING.rst
medubelko: `CONTRIBUTING` is preferred, but I don't think we should tackle that right now.

.readthedocs.yaml
sergiusens: warnings,
```
WARNING: The local Git repository could not be found.
```

Seems like craft-cli had these once upon a time (not longer than a month ago), so I decided to allow them for now https://app.readthedocs.com/projects/canonical-craft-cli/builds/2700288/#30829517--59

sergiusens (approved):

PR #73, opened 2025-01-20
feat: Implement Volumes specification subset

upils:
Implement the subset of the Volumes specification defined in https://docs.google.com/document/d/12LxcUIUEoXvzK619CvK9eHU_8n_rEKnLPEC4clsriXA/edit?pli=1&tab=t.0

CRAFT-3972

### DIFF
--- a/imagecraft/models/project.py
+++ b/imagecraft/models/project.py
@@ -1,6 +1,6 @@
 # This file is part of imagecraft.
 #
-# Copyright 2023 Canonical Ltd.
+# Copyright 2023-2025 Canonical Ltd.
 #
 # This program is free software: you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 3, as published
@@ -26,6 +26,7 @@ from craft_application.errors import CraftValidationError
 from craft_application.models import BuildPlanner as BaseBuildPlanner
 from craft_application.models import Platform as BasePlatform
 from craft_application.models import Project as BaseProject
+from craft_application.models.constraints import SingleEntryDict
 from craft_providers import bases
 from pydantic import (
     ConfigDict,
@@ -34,6 +35,11 @@ from pydantic import (
     model_validator,
 )
 from typing_extensions import override
+
+from imagecraft.models.volume import (
+    Volume,
+    VolumeName,
+)
 
 
 class Platform(BasePlatform):
@@ -94,6 +100,7 @@ class Project(BaseProject):
     base: BaseT  # type: ignore[reportIncompatibleVariableOverride]
     build_base: BuildBaseT  # type: ignore[reportIncompatibleVariableOverride]
     platforms: dict[str, Platform] | None = None  # type: ignore[assignment, reportIncompatibleVariableOverride]
+    volumes: SingleEntryDict[VolumeName, Volume]
 
     model_config = ConfigDict(
         validate_assignment=True,
--- a/imagecraft/models/volume.py
+++ b/imagecraft/models/volume.py
@@ -117,10 +117,7 @@ StructureList = UniqueList[StructureItem]
 class Volume(CraftBaseModel):
     """Volume defining properties of the image."""
 
-    volume_schema: typing.Annotated[
-        Literal[PartitionSchema.GPT],
-        Field(default=PartitionSchema.GPT),
-    ] = Field(alias="schema")
+    volume_schema: Literal[PartitionSchema.GPT] = Field(alias="schema")
     structure: StructureList = Field(min_length=1)
 
     @field_validator("structure", mode="after")
--- a/imagecraft/models/volume.py
+++ b/imagecraft/models/volume.py
@@ -1,0 +1,137 @@
+# This file is part of imagecraft.
+#
+# Copyright 2025 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 3, as published
+# by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranties of MERCHANTABILITY,
+# SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Volume configuration pydantic model."""
+
+import collections
+import enum
+import re
+import typing
+import uuid
+from typing import Literal, Self, cast
+
+from craft_application.models import (
+    CraftBaseModel,
+)
+from craft_application.models.constraints import (
+    UniqueList,
+    get_validator_by_regex,
+)
+from pydantic import (
+    BeforeValidator,
+    ByteSize,
+    Field,
+    StringConstraints,
+    field_validator,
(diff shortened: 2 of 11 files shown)


imagecraft/models/volume.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Role(str, enum.Enum):
+    """Role describes the role of given structure."""
+
+    SYSTEM_DATA = "system-data"
+    SYSTEM_BOOT = "system-boot"
+
+
+class StructureItem(CraftBaseModel):
```
upils, line 60: Hopefully this name helps avoid the confusion with the "structure" field in the Volume class. In snapd it is named "VolumeStructure" but I am not sure that is better.

imagecraft/models/volume.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class Volume(CraftBaseModel):
+    """Volume defining properties of the image."""
+
+    volume_schema: typing.Annotated[
+        Literal[PartitionSchema.GPT],
+        Field(default=PartitionSchema.GPT),
+    ] = Field(alias="schema")
+    structure: UniqueList[StructureItem] = Field(min_length=1)
```
upils, line 94: I am tempted to rename it "structures" (plural) and only expose the singular name to the user. I would make it clear in the code it holds a list of "structure items".

imagecraft/models/project.py
```diff
@@ -112,3 +121,13 @@ def _providers_base(cls, base: str) -> bases.BaseAlias | None:
         :returns: Always None because imagecraft only supports bare bases.
         """
         return None
+
+    @field_validator("volumes", mode="after")
+    @classmethod
+    def _validate_volumes(
```
upils, line 127: It seems pydantic it not validating dictionary **keys** even with the `VolumeName` annotation set above, so the next best thing I thought off was to add this validator. I would be interested to know if there is a better way.

imagecraft/models/project.py
```diff
@@ -94,6 +102,7 @@ class Project(BaseProject):
     base: BaseT  # type: ignore[reportIncompatibleVariableOverride]
     build_base: BuildBaseT  # type: ignore[reportIncompatibleVariableOverride]
     platforms: dict[str, Platform] | None = None  # type: ignore[assignment, reportIncompatibleVariableOverride]
+    volumes: SingleEntryDict[VolumeName, Volume]
```
tigarmo, line 105: a SingleEntryDict is a dict with a single key; the spec says that `volumes` can have multiple items, is this here on purpose?

imagecraft/models/project.py
tigarmo: interesting; maybe the definition of VolumeName needs to be more similar to [ProjectName](https://github.com/canonical/craft-application/blob/main/craft_application/models/constraints.py#L102)

imagecraft/models/volume.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class StructureItem(CraftBaseModel):
+    """Structure item of the image."""
+
+    name: StructureName
+    id: uuid.UUID | None = None
+    role: Role
+    structure_type: GptType = Field(alias="type")
+    size: ByteSize = ByteSize(0)
```
tigarmo, line 74: out of curiosity, what is the semantics of size 0?

imagecraft/models/project.py
upils: The original Volumes spec says that, but the subset of this spec that is defined in the specification you reviewed is stating that a single entry is allowed (this will change in the future but for now we support a single one).

imagecraft/models/volume.py
upils: I was unsure about the default value, because if no size is set in the configuration imagecraft must create a partition of the size of the content that will be set in it. In the spec I mentioned "Defaults to the size of the content set in the structure." on the `size` key.

Should I set `size: ByteSize | None = None` instead?

imagecraft/models/project.py
upils: Thanks, this is cleaner. The only downside is that I needed a strict match (`re.fullmatch()`) so for now I had to create a more strict version of the `VALID_PARTITION_REGEX` regex.

imagecraft/models/volume.py
tigarmo: no, I think 0 is fine. I just haven't fully read the spec.
Might want to add a docstring to this field describing that if `size is 0`, it's computed from the contents

tigarmo (approved):
thanks!

imagecraft/models/volume.py
```diff
@@ -0,0 +1,137 @@
...
+
+# This is slightly stricter than VALID_PARTITION_REGEX from craft_parts.utils.partition_utils.
+PARTITION_COMPILED_REGEX = re.compile(r"^(?!-|\.)[a-z0-9-\.]+(?<!-|\.)$", re.ASCII)
+
+VOLUME_NAME_COMPILED_REGEX = PARTITION_COMPILED_REGEX
+STRUCTURE_NAME_COMPILED_REGEX = PARTITION_COMPILED_REGEX
+
+VOLUME_INVALID_MSG = (
+    "volume names must only contain lowercase letters, numbers, "
+    "and hyphens, and may not begin or end with a hyphen."
```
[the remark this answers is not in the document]
upils: After extending and refactoring the collision detection (see https://github.com/canonical/craft-parts/compare/main...work/allow-separator-in-partition-names) every cases I could think off are correctly detected.
Allowing hyphens in volume names will make collisions more likely but I am not sure that is a big issue. Moreover we never parse a partition name to get the volume name or structure name (just like component names are not parsed from a `component/name` partition) so the place it could be problematic is when dealing with env vars, and in that case the collision detection (running when validating partition names) make sure this is fine.

Would you mind taking a look at the prototype in craft-parts and see if it looks safe enough?

imagecraft/models/volume.py
```diff
@@ -0,0 +1,137 @@
...
+    Field,
+    StringConstraints,
+    field_validator,
+    model_validator,
+)
+
+from imagecraft.platforms import GPT_NAME_MAX_LENGTH, FileSystem, GptType
+
+# This is slightly stricter than VALID_PARTITION_REGEX from craft_parts.utils.partition_utils.
+PARTITION_COMPILED_REGEX = re.compile(r"^(?!-|\.)[a-z0-9-\.]+(?<!-|\.)$", re.ASCII)
```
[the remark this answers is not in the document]
upils: It is the craft-parts partitions. In the validation here we need a exact match so I needed to add "^" at the begining and "$" at the end (because I did not find a way to tell pydantic the pattern must be used to validate strictly). But   I found a better way to express that (using the original pattern) so the difference will be obvious.

imagecraft/models/volume.py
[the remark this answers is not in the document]
cmatsuoka: Craft-parts has the partitions feature operating under a certain set of rules and I don't think we should modify that to accommodate this use case. Imagecraft can handle all necessary translations on the application side, it just needs to translate `volume/disk1/name` to namespaced partition `disk1_name` under the `volume` namespace, and act exactly as snapcraft does with components.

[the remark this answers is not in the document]
cmatsuoka: Discussed with @upils and @mr-cal and agreed that the proposed way is the best solution.

cmatsuoka (approved):
LGTM, thanks!

PR #86, opened 2025-02-03
feat: add bash completion script

bepri:
- [ ] Have you successfully run `tox`?

---

CRAFT-3878

### DIFF
--- a/imagecraft/cli.py
+++ b/imagecraft/cli.py
@@ -16,6 +16,10 @@
 
 """Command-line application entry point."""
 
+from typing import Any
+
+from craft_cli import Dispatcher
+
 from imagecraft.application import Imagecraft
 from imagecraft.services.service_factory import ImagecraftServiceFactory
 
@@ -38,3 +42,11 @@ def _create_app() -> Imagecraft:
     )
 
     return Imagecraft(app=APP_METADATA, services=services)
+
+
+def get_app_info() -> tuple[Dispatcher, dict[str, Any]]:
+    """Retrieve application info. Used by craft-cli's completion module."""
+    app = _create_app()
+    dispatcher = app._create_dispatcher()  # noqa: SLF001 (private member access)  # pyright: ignore[reportPrivateUsage]
+
+    return dispatcher, app.app_config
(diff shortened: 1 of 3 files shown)


imagecraft/cli.py
```diff
@@ -38,3 +42,11 @@ def _create_app() -> Imagecraft:
     )
 
     return Imagecraft(app=APP_METADATA, services=services)
+
+
+def get_app_info() -> tuple[Dispatcher, dict[str, Any]]:
+    """Retrieve application info. Used by craft-cli's completion module."""
+    app = _create_app()
+    dispatcher = app._create_dispatcher()  # noqa: SLF001 (private member access)  # pyright: ignore[reportPrivateUsage]
```
bepri, line 50: Unfortunately I couldn't figure out a way to make both Pyright and Ruff happy with this one... I've gotta access this private member due to the current implementation of the completion generator - this could be fixed if we just made a public method but that seemed excessive for this one use.

lengau: `create_dispatcher` should probably be public at this point in craft-application, but that's out of scope here.

lengau (approved):

upils (changes requested):
Thanks! It looks good. 2 small notes though:
- the git history looks weird. Could you rebase on main so you do not have the merge commit?
- I see we did not write spread tests for this in any other *craft application. Was it mentioned at some point? (maybe I missed something). I tested by hand but it is hard to make sure it will not break in the future, but I guess that is something hard to automate.

bepri:
>     * the git history is looks weird. Could you rebase on main so you do not have the merge commit?
Done!
>     * I see we did not write spread tests for this in any other *craft application. Was it mentioned at some point? (maybe I missed something). I tested by hand but it is hard to make sure it will not break in the future, but I guess that is something hard to automate.
Correct. We initially wanted some integration tests on the lib side to simulate <TAB> presses, but this seemed really difficult and infeasible to do. Some research showed that tests for tab completion scripts aren't terribly standard anyways, so the team decided to settle on "by-hand" testing. The only currently known issue is that it cannot tab-complete global arguments (CRAFT-4007), so as long as everything else seems to work, we've been merging it.

upils (approved):

PR #88, opened 2025-02-04
Create craft partitions from structure items

upils:

---

Implement the dynamic craft partitions creation based on structures in the Volumes section.


pyproject.toml
```diff
@@ -5,7 +5,7 @@ dynamic = ["version", "readme"]
 authors = [{ name = "Canonical Ltd.", email = "snapcraft@lists.snapcraft.io" }]
 license = { file = "LICENSE" }
 dependencies = [
-    "craft-parts~=2.2",
+    "craft-parts @ git+https://github.com/canonical/craft-parts.git@work/CRAFT-4064-Allow-slashes-in-partition-names",
```
upils, line 8: Reminder: use craft-parts new release when this branch is merged.

imagecraft/models/project.py
```diff
@@ -94,13 +95,16 @@ def _preprocess_all_platforms(cls, platforms: dict[str, Any]) -> dict[str, Any]:
         return platforms
 
 
+VolumeDictT = SingleEntryDict[VolumeName, Volume]
```
cmatsuoka, line 98: I suppose we're using this type as a shortcut because we only allow one volume. Is the error message reasonable if someone tries to define more than one? I think the best way to implement this would be to make it a normal dict and fail in validation if there's more than one instead of (ab)using the existing type. @tigarmo wdyt?

mr-cal (approved):
Nice to see the integration of partitions for imagecraft is minimal.

imagecraft/models/project.py
tigarmo: I agree; I think you can also do something like `VolumeDictT = Annotated[dict, Field(min_length=1, max_length=1)]` and see what the error message looks like.

upils: This is what I did first and then saw the SingleEntryDict type. What is the purpose of this type if not to be used in my case? 

Using this type the error is `- dictionary should have at most 1 item after validation, not 2 (in field 'volumes')`  (tested in `tests/unit/models/test_project.py`) which looks reasonable to me. @tigarmo with your suggestion the error is the same because you suggested almost exactly the same definition as `SingleEntryDict`:

```
SingleEntryDict = Annotated[
    dict[T, Tv],
    pydantic.Field(min_length=1, max_length=1),
]
```

tigarmo: > @tigarmo with your suggestion the error is the same because you suggested almost exactly the same definition as `SingleEntryDict`:

Ah my bad, I didn't know this.

> Using this type the error is - `dictionary should have at most 1 item after validation, not 2 (in field 'volumes')`

The message is a bit cryptic but I'm fine with it either way.

cmatsuoka: I'd be ok with a new definition for `VolumeT` (because later if we accept more entries we would be changing type constraints and not the type itself). The ideal case would be a proper validation error message telling that we currently support only a single volume, and we can just remove this validation later if we decide to suport multi-volumes.

tigarmo (approved):

PR #94, opened 2025-02-21
feat: disk creation

mattculler:

---

Implement a functional package service, as defined by ST142.  Service will create images, but they won't be populated due to currently unimplemented organizing from overlays into partitions.

(CRAFT-4154)


mattculler:
The new `pack/diskutil.py` and `pack/gptutil.py` files were copied in from another private project - some of the functions there are unused, but I'm leaving them for now because they might be used in CRAFT-4155 (make imagecraft artifacts bootable).

tests/unit/services/test_pack.py
```diff
@@ -13,11 +13,31 @@
...
+
+    def mocked_open(self, *args, **kwargs):
+        if self.name == "testo.txt":
+            return opener(self, *args, **kwargs)
+        print("open", self, args, kwargs)
+        return open(*args)  # noqa: PTH123
+
+    with (
+        patch("imagecraft.services.pack.diskutil", autospec=True) as diskutil,
+        patch("imagecraft.services.pack.gptutil", autospec=True) as gptutil,
```
sergiusens, line 36: why not mocker? Also remove the need for importing mock_open https://pytest-mock.readthedocs.io/en/latest/usage.html

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,319 @@
...
+        f"if={str(partition)}",
+        f"of={str(imagepath)}",
+        f"bs={sector_size}",
+        f"seek={sector_offset}",
+        f"count={sector_count}",
+        "conv=notrunc,sparse",
+    )
+
+
+def compare_contents_partition_size(
```
sergiusens, line 300: I do not see this used anywhere

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,319 @@
...
+    """Convert KiB to sector count.
+
+    The sector count will be rounded up to the nearest sector boundary
+
+    :param gibibyte: Kibibytes.
+    :param sector_size: Size of a sector.
+
+    :returns: Number of sectors.
+    """
+    return bytes_to_sectors(kibibyte, sector_size, _unit_multiplier=KIB)
```
sergiusens, line 73: I see no usage of this in the code

imagecraft/subprocesses.py
```diff
@@ -0,0 +1,40 @@
...
+"""Imagecraft subprocess utility functions."""
+
+import subprocess
+from subprocess import PIPE, CompletedProcess
+from typing import Any, cast
+
+from craft_cli import emit
+
+
+def run(cmd: str, *args: Any, **kwargs: Any) -> CompletedProcess[str]:
```
lengau, line 24: This might be nice to have upstreamed somewhere in a more general way. [craft-parts](https://github.com/canonical/craft-parts/blob/37cafc6cfd8c12afb89847fe36c409a20975e1ed/craft_parts/utils/os_utils.py#L327-L345) does a version of this prepending `":: "` to each line of output in a log (but doesn't log the command itself) that could probably be merged into this in some way, and it would be better to have some version of this for all the other places we use subprocess in our apps.

imagecraft/subprocesses.py
```diff
@@ -0,0 +1,40 @@
...
+    defaults = {
+        "text": True,
+        "check": True,
+        "stdout": PIPE,
+    }
+    for key, value in defaults.items():
+        if key not in kwargs:
+            kwargs[key] = value
+
+    emit.debug(f"Command: {cmd} {' '.join([str(a) for a in args])}")
```
lengau, line 36: Two improvements here:

1. `shlex.join` will lex it so tokens that contain spaces are escaped and quoted. Probably not a problem for current usage of this function, but once it exists we may use it for more!
2. Having a list as an intermediate object from this generator is unnecessary.


```suggestion
    emit.debug(f"Command: {cmd} {'shlex.join(str(a) for a in args)}")
```

imagecraft/subprocesses.py
```diff
@@ -0,0 +1,40 @@
...
+    return cast(
+        CompletedProcess[str],
+        subprocess.run([cmd] + [str(a) for a in args], **kwargs),  # noqa: PLW1510
```
lengau, line 39: Likewise here we can create just a single list using the command and a generator expression.

```suggestion
        subprocess.run([cmd, *(str(a) for a in args)], **kwargs),  # noqa: PLW1510
```

pyproject.toml
```diff
@@ -355,3 +355,6 @@ classmethod-decorators = ["pydantic.validator", "pydantic.root_validator"]
     "I001", # isort leaves init files alone by default, this makes ruff ignore them too.
     "F401", # Allows unused imports in __init__ files.
 ]
+"tools/imgmount.py" = [
+    "PLR2004", # Various invocations can have variable numbers of args, magic variable ok
```
lengau, line 359: When targeted at a specific file like this, it's probably better to have this in the file itself. At the top of the file you can add:

```python
# Various invocations here can have multiple numbers of args, so magic values are ok.
# ruff: noqa: PLR2004
```

tests/unit/services/test_pack.py
lengau: I personally tend to avoid `mocker` because the way it stacks its patches can be weird and counterintuitive when combined with other tools, but in this case I think it makes more sense.

tools/imgmount.py
lengau, line 1: This is more a question for @upils than for you:

Would this be worth having inside the imagecraft package, possibly with its own entrypoint? It might be nice in the snap to have `imagecraft.mount` and `imagecraft.umount` commands as separate entrypoints.

tests/conftest.py
```diff
@@ -135,7 +135,13 @@ def default_factory(default_project, default_build_plan):
         project=default_project,
     )
 
-    factory.update_kwargs("lifecycle", build_plan=default_build_plan)
+    factory.update_kwargs(
+        "lifecycle",
+        work_dir=Path("work/"),
```
lengau, line 140: These really should use a `tmp_path`

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,319 @@
...
+
+def create_zero_image(*, imagepath: Path, sector_size: int, sector_count: int) -> None:
+    """Create an empty image.
+
+    :param imagepath: Path to image file.
+    :param sector_size: Size of a sector.
+    :param sector_count: Number of sectors.
+    """
+    run(
+        "dd",
```
lengau, line 104: If we're using this we probably want to ensure that the `$PATH` in the snap includes the bin directory from the base snap so we're using the base snap's `dd` rather than the system's `dd`. In this particular case I don't expect it to be an issue, but if we come along and change this command at some point to use a new feature from core26 dd or something, we don't want it to fail on older systems.

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,319 @@
...
+
+    if label is not None:
+        mke2fs_args.extend(["-L", label])
+
+    if uuid is not None:
+        mke2fs_args.extend(["-U", uuid])
+
+    mke2fs_args.append(partitionpath)
+
+    run("mke2fs", *mke2fs_args)
```
lengau, line 155: Similar here, but with the base snap's `/sbin`

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def format_install_fat_partition(  # pylint: disable=too-many-arguments
+    *,
+    fattype: str,
```
lengau, line 160: `format_install_fat_partition(fattype="btrfs", ...`

lengau:
```suggestion
    fattype: Literal["fat", "vfat"],
```

imagecraft/pack/diskutil.py
mattculler: For some reason I was thinking that `grubutil.py` which will pull in to imagecraft for CRAFT-4155 used these, but it doesn't - removing.

imagecraft/pack/diskutil.py
mattculler: Removed.

imagecraft/services/pack.py
```diff
@@ -43,15 +49,75 @@ def __init__(
...
+            sector_size=SECTOR_SIZE,
+            sector_count=image_sectors,
+        )
+
+        # Create partition layout
+        gptutil.create_gpt_layout(
+            imagepath=disk_image_file,
+            sector_size=SECTOR_SIZE,
+            layout=volume,
+        )
```
lengau, line 81: This seems like something we'd want to put in its own helper function, both for the ability to test this section on its own and so it's easier to replace if we ended up with something like...

```python
if we_want_gpt():
    create_gpt_image(disk_image_file, volume)
else:
    create_fancy_futuristic_image(disk_image_file, volume)
```

imagecraft/services/pack.py
```diff
@@ -43,15 +49,75 @@ def __init__(
...
+
+        # Create filesystems
+        project_dirs = self._services.lifecycle.project_info.dirs
+        with tempfile.TemporaryDirectory() as partition_dir:
+            for structure_item in volume.structure:
+                partition_name = f"volume/{volume_name}/{structure_item.name}"
+                partition_prime_dir = project_dirs.get_prime_dir(
+                    partition=partition_name
+                )
```
lengau, line 90: If we haven't modified a partition, do we really need to recreate it? Since this is running inside the managed container, we could keep these files in their own directory somewhere and only recreate them if that partition's prime directory has been touched.

imagecraft/services/pack.py
```diff
@@ -43,15 +49,75 @@ def __init__(
...
+
+                # Test data - remove once we can organize from overlays into partitions
```
lengau, line 92: Do you have an issue you could link to here?

pyproject.toml
mattculler: Nice, that's the ignore syntax I was looking for but didn't find, fixed.

tools/imgmount.py
mattculler: That could totally make sense, it's a good idea.  This file started as a little shell script and then I got carried away and accidentally made it awesome, I'll let Paul decide if its functionality should be incorporated into imagecraft proper.

imagecraft/services/pack.py
mattculler: I tried organizing into partitions and it works!  So this was removed.

imagecraft/services/pack.py
mattculler: Conversed OOB, added link to issue that would conflict with Alex's suggestion.

imagecraft/pack/diskutil.py
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/6bfa4ee0f123ec9ca41058482aab2fc3d0f7b878

tests/unit/services/test_pack.py
mattculler: Fixed

imagecraft/subprocesses.py
mattculler: Issue opened https://github.com/canonical/craft-parts/issues/1023

tools/imgmount.py
lengau: Yeah, the actual change is also probably out of scope here but worth considering for later.

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,263 @@
...
+    sector_size: int,
+    unit_multiplier: int = 1,
+) -> int:
+    """Convert bytes to sector count.
+
+    The sector count will be rounded up to the nearest sector boundary
+
+    :param bytes_: Byte count.
+    :param sector_size: Size of a sector.
+    :param unit_multiplier: Use one of the *IB constants here if the input unit is
```
lengau, line 50: If this were a library I'd say you should make those values part of an enum and then make sure that enum type is used here, but that might be micro-optimising our types

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,263 @@
...
+
+    if label is not None:
+        mke2fs_args.extend(["-L", label])
+
+    if uuid is not None:
+        mke2fs_args.extend(["-U", uuid])
+
+    mke2fs_args.append(partitionpath)
+
+    run("mke2fs", *mke2fs_args)
```
lengau, line 120: Probably no harm in returning the `CompletedProcess` here in case anything needs to examine it.

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,263 @@
...
+    """Format partition EXT2/3/4 and copy files.
+
+    :param fstype: Type of Ext filesystem (ext2/3/4).
+    :param content_dir: Directory containing contents for partition.
+    :param partitionpath: Path to partition file.
+    :param sector_size: Size of a sector.
+    :param sector_count: Number of sectors.
+    :param label: Ext Filesystem label, empty if not supplied.
+    :param uuid: Ext Filesystem UUID, generated if not supplied.
+    """
```
lengau, line 97: Maybe note the types of errors this could raise - I think they're all just a SubprocessError, but in future we might want to catch and parse those.

imagecraft/pack/diskutil.py
mattculler: It's a good idea.  I had been eyeing ways to avoid passing sector size everywhere, I think maybe the way to go is a dataclass that takes sector size + (Gi/Mi/Ki/G/M/K/bytes) and can return whatever unit a caller needs.

imagecraft/pack/gptutil.py
```diff
@@ -0,0 +1,224 @@
...
+
+from craft_cli import CraftError, emit
+
+from imagecraft.models import Volume
+from imagecraft.pack import diskutil
+from imagecraft.subprocesses import run
+
+# Supported Sector Sizes
+
+SUPPORTED_SECTOR_SIZES = [512]
```
lengau, line 29: Not strictly necessary, but I like seeing constants be immutable types as it prevents accidental mutation.

```suggestion
SUPPORTED_SECTOR_SIZES = (512,)
```

imagecraft/pack/diskutil.py
mattculler: I think if a caller needed to examine the `CompletedProcess` we'd want to figure out why and return something more specific/structured.  The function will already raise if `mke2fs` exits nonzero.  And there's a `CompletedProcess` that's not being returned by the `create_zero_image`/`dd` call too.  Hopefully we can keep `CompletedProcess` as an abstraction that's deeper than the level of this interface.

imagecraft/pack/diskutil.py
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/13f2ab03efbb06dc00c7d9b08b74a25a50929267

tools/imgmount.py
mattculler: https://github.com/canonical/imagecraft/issues/95

imagecraft/services/pack.py
```diff
@@ -43,15 +49,60 @@ def __init__(
         self._build_plan = build_plan
 
     @override
-    def pack(self, prime_dir: pathlib.Path, dest: pathlib.Path) -> list[pathlib.Path]:  # noqa: ARG002
+    def pack(self, prime_dir: Path, dest: Path) -> list[Path]:  # noqa: ARG002
```
lengau, line 52: Not about your code here, but about how craft-application assumes we're packing one prime directory. I'd like thoughts on this - one thing is maybe to expose `get_prime_dir` directly in the lifecycle service so we can use it from the package service without specifically pulling out the project info.

Maybe we should also give a way to iterate over the prime dirs for each partitions? Something like:

```python
for partition_name, prime_dir in self._services.lifecycle.get_prime_dirs():
  # Do stuff
```

lengau (approved):

imagecraft/services/pack.py
mattculler: Very much agreed that the signature for `pack` is incorrect for applications using partitions.  But, the current signature _is_ correct for applications _not_ using partitions.

Do we need two pack methods?  Like:
```
    def pack(self, prime_dir: Path, dest: Path) -> list[Path]
    def pack_partitions(self, prime_dirs: dict[str, Path], dest: Path) -> list[Path]
```

This seems most flexible - whether the application wants to directly pull partition prime dirs by name (like imagecraft and the private application I maintain currently do) or iterate like you show above.

tests/unit/services/test_pack.py
upils, line 1: I think we should add more tests to cover this "major" feature. Suggestion:

- extend `test_imagecraft_build` (and/or rename it?) to call `pack` and check the resulting image layout and content.
- unit test `pack` with a variety of structures
- add a test to make sure the content is never overriding the GPT backup partition table (at the end of the disk). We had this kind of issue in ubuntu-image and I would like to make sure we get a proper design preventing this right from the beginning.
- since imagecraft is now able to actually build an image I think we can add a first real spread test too.

imagecraft/services/pack.py
```diff
@@ -43,15 +49,60 @@ def __init__(
...
+                diskutil.inject_partition_into_image(
+                    partition=partition_img,
+                    imagepath=disk_image_file,
+                    sector_size=SECTOR_SIZE,
+                    sector_offset=gptutil.get_partition_sector_offset(
+                        disk_image_file,
+                        structure_item.name,
+                    ),
+                    sector_count=sector_count,
+                )
```
upils, line 104: suggestion: Before returning, what do you think of adding a sanity check with sfdisk to make sure the resulting image has a valid partition layout? Ideally that should not be necessary but that might help catch a invalid disk before is it copied to the target system.

imagecraft/pack/diskutil.py
upils, line 1: Initially I thought this file and `gptutils.py` would live in the `platforms` directory. I think it make sense to either move them there or merge the ones in `platforms` with the ones in `pack`. What do you think?

upils (changes requested):
Thanks @mattculler! Great work!

I also think having to pass around `sector_size` and `sector_count` is a bit weird and feels like a leaking implementation details. If we were not using `dd` under the hood we would not have to care about that. What do you think of passing the `sector_size` and the `size` instead? (and get the sector count at the last moment, before calling `dd`)? I think passing around the partition/disk size is making more sense.

imagecraft/pack/gptutil.py
```diff
@@ -0,0 +1,238 @@
...
+) -> None:
+    """Create a zeroed image file with a GPT partition table, but no filesystems or data.
+
+    :param imagepath: Path to image file.
+    :param sector_size: Sector size in bytes.
+    :param layout: Disk layout to create.
+    """
+    # Determine necessary image size, and reserve space
+    # "The default start offset for the first partition is 1 MiB", per `man sfdisk`,
+    # plus one more MiB for padding at the end of the disk.
```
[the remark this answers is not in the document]
mattculler: I had put that in there just because it made it work, but your comment got me thinking so I played around with it.  What we actually need are: 1MB for the start of disk reserved space, +1 MB for empty space between each pair of adjacent partitions.  The reason for this is because partitions can end wherever the end happens to fall, but the start of the next partition needs to be aligned to a sector boundary.  So the calculation is `1+ (len(partitions) - 1)`.

https://github.com/canonical/imagecraft/pull/94/commits/618961ac6bee5dc0fcd4f53f0d8642bdb3573afb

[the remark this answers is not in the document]
mattculler: OOB conversation with @upils - end of disk padding is to avoid overwriting the backup GPT header.

imagecraft/pack/diskutil.py
mattculler: Imports became circular when merging the `platforms/*util` stuff into `pack/*util` (`volume.py` needs `GptType` and `FileSystem`, `gptutil` needs `Volume`) so I put the stuff in `platforms` into `volume.py`.

https://github.com/canonical/imagecraft/pull/94/commits/af97066798d8e16bdf075c6788c18452a3925796

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+        "dd",
+        "if=/dev/zero",
+        f"of={str(imagepath)}",
+        f"bs={sector_size}",
+        f"count={sector_count}",
+        "conv=sparse",
+    )
+
+
+def format_install_ext_partition(  # pylint: disable=too-many-arguments
```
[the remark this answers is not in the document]
mattculler: Yeah you're right.  I like `format_populate_xxx_partition`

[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/b1ed6d974cbb9d75a04ffb7b5322b1da142cf6ef

imagecraft/services/pack.py
```diff
@@ -43,15 +49,60 @@ def __init__(
...
+                    partitionpath=partition_img,
+                    sector_size=SECTOR_SIZE,
+                    sector_count=sector_count,
+                    label=structure_item.filesystem_label,
+                )
+                diskutil.inject_partition_into_image(
+                    partition=partition_img,
+                    imagepath=disk_image_file,
+                    sector_size=SECTOR_SIZE,
+                    sector_offset=gptutil.get_partition_sector_offset(
```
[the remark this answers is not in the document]
mattculler: I think that's a good idea, but to do it cleanly my thought would be to add state to gptutil rather than hanging on to that dict in the caller.  And that would be more refactoring than I'd like to do here to save a few subprocesses.  Should we open an issue for this?

imagecraft/pack/gptutil.py
```diff
@@ -0,0 +1,238 @@
...
+            "name": f'"{structure_item.name}"',
+            "size": str(
+                diskutil.bytes_to_sectors(
+                    structure_item.size,
+                    sector_size,
+                )
+            ),
+            "type": structure_item.structure_type.value,
+        }
+        if structure_item.role == "system-boot":
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/0811f635f384d068cbc59e69bb887141a123e0c5

imagecraft/services/pack.py
```diff
@@ -43,15 +49,60 @@ def __init__(
...
+            layout=volume,
+        )
+
+        # Create partition images with filesystems.  These are always recreated, but we
+        # may want to revisit that once this is solved:
+        # https://github.com/canonical/craft-parts/issues/665
+        project_dirs = self._services.lifecycle.project_info.dirs
+        with tempfile.TemporaryDirectory() as partition_dir:
+            for structure_item in volume.structure:
+                partition_name = f"volume/{volume_name}/{structure_item.name}"
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/f304fdcdc035db33cb260e44824371c7a08f21e0

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+    *,
+    fstype: str,
+    content_dir: Path,
+    partitionpath: Path,
+    sector_size: int,
+    sector_count: int,
+    label: str | None = None,
+    uuid: str | None = None,
+) -> None:
+    """Format partition EXT2/3/4 and copy files.
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/dcbbe7d3908ce30ea6727c7a0f63871acad843a7

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+    :raises CalledProcessError: If dd or mke2fs fails.
+    """
+    # Create the partition file
+    create_zero_image(
+        imagepath=partitionpath, sector_size=sector_size, sector_count=sector_count
+    )
+
+    # Create and copy
+    mke2fs_args = [
+        "-Eno_copy_xattrs",
```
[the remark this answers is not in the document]
mattculler: That's a good question.  It came from the private project I copied this code from.  Possibly before that it originally came from craft-parts and was part of the xattr stripping we do before final artifact packing?

In any case we shouldn't need it for imagecraft.

https://github.com/canonical/imagecraft/pull/94/commits/96c232a491899c50f45d427901282402be69ea80

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+    :param uuid: Fat Filesystem UUID, generated if not supplied.
+    :raises CalledProcessError: If dd, mkfs.xxx, or mcopy fails.
+    """
+    # Create the partition file
+    create_zero_image(
+        imagepath=partitionpath, sector_size=sector_size, sector_count=sector_count
+    )
+
+    # Create and copy
+    mkdosfs_args: list[str | Path] = []
```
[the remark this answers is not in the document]
mattculler:
```
$ realpath `which mkdosfs`
/usr/sbin/mkfs.fat
$ realpath `which mkfs.fat`
/usr/sbin/mkfs.fat
$ realpath `which mkfs.vfat`
/usr/sbin/mkfs.fat
```

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+
+    run("mkfs." + fattype, *mkdosfs_args)
+
+    if any(content_dir.iterdir()):
+        # If we invoke mcopy directly, the sh wrapper will quote the
+        # source path because it contains a wildcard. This will confuse
+        # mcopy. Instead, we wrap the call in bash to get it to
+        # remove the quotes. Mcopy will fail if the content directory is
+        # empty.
+        # Note that the -i flag to mcopy seems to be completely undocumented.
```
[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/ea7c0a6460d6ed912151e0ecbaeb4d902414bd95

imagecraft/pack/gptutil.py
```diff
@@ -0,0 +1,238 @@
...
+    layout: Volume,
+) -> None:
+    """Partition image.
+
+    :param imagepath: Path to image file.
+    :param sector_size: Sector size in bytes.
+    :param layout: Disk layout to create.
+    :raises CalledProcessError: If sfdisk fails.
+    """
+    # There is no way today to specify the sector size when using
```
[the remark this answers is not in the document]
mattculler: Do you mean the `sector-size` header field in `sfdisk`?

This comment came from the private repo, I think what it means is that when we're making an image file, we won't know what the physical sector size is for some hypothetical eventual physical media.

I don't think this comment adds much, especially in the context of imagecraft.  Removing, but pasting here for posterity:

>     # There is no way today to specify the sector size when using
    # sfdisk (or any other variants) directly on an image file as
    # the block device is not present without a loopback mount. We
    # cannot use loopback devices on unprivileged containers, and
    # even with the latter enabled, loopback control is not
    # namespaced and therefore not elegant or safe to use.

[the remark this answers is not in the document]
mattculler: Also man sfdisk has this to say, which might be related to this comment:

> sector-size
           Specify sector size. sfdisk always uses device sector size. Since version 2.39 sfdisk recalculates sizes from dump if the script and device sector size differ.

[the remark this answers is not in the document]
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/29c481d17e7d5ff37112501b2b9ef94a61c4f79c

pyproject.toml
```diff
@@ -56,12 +56,18 @@ docs = [
     "sphinxcontrib-details-directive",
     "matplotlib",
 ]
+lint = [
```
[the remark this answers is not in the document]
mattculler: I thought the workflow calling `--extra lint` was in starflow, but it's actually here.  Fixed: 

https://github.com/canonical/imagecraft/pull/94/commits/c55d67200e61c775024309174522b0e0e77f75d

imagecraft/services/pack.py
```diff
@@ -14,19 +14,25 @@
...
 
 from imagecraft.models import Project
+from imagecraft.pack import diskutil, gptutil
 
-if typing.TYPE_CHECKING:
+if TYPE_CHECKING:
     from imagecraft.services import ImagecraftServiceFactory
 
 
+SECTOR_SIZE = 512
```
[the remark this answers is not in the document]
mattculler: This is more or less a placeholder to avoid hardcoding 512 byte sectors all over the place - just hardcode it in one place!  Later we can decide what to do about the sector size used (user selectable, based on physical media, etc).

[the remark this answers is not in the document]
mattculler: Available sector sizes are here: https://github.com/canonical/imagecraft/pull/94/files#diff-989b11645780fc9e43a5ced57502d785cd2401f41b7d143988df62566109b8a0R30

imagecraft/pack/diskutil.py
```diff
@@ -0,0 +1,267 @@
...
+    :param sector_count: Number of sectors.
+    :raises CalledProcessError: If dd fails.
+    """
+    run(
+        "dd",
+        "if=/dev/zero",
+        f"of={str(imagepath)}",
+        f"bs={sector_size}",
+        f"count={sector_count}",
+        "conv=sparse",
```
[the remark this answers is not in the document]
mattculler: Those commands might be faster (especially `fallocate`), but that `dd conv=sparse` directive I have probably means they're all about the same.  This dd call isn't noticeably slow on the yamls I've tested.  Perhaps benchmarking alternatives should be another task.

mattculler:
> Thanks @mattculler! Great work!
> 
> I also think having to pass around `sector_size` and `sector_count` is a bit weird and feels like a leaking implementation details. If we were not using `dd` under the hood we would not have to care about that. What do you think of passing the `sector_size` and the `size` instead? (and get the sector count at the last moment, before calling `dd`)? I think passing around the partition/disk size is making more sense.

https://github.com/canonical/imagecraft/pull/94/commits/76b591a83bc3246ca84e0d3020beeba27fe51161

imagecraft/services/pack.py
mattculler: https://github.com/canonical/imagecraft/pull/94/commits/d5e1bc77a1245f018e17d5930fb1605f06c83d65

imagecraft/services/pack.py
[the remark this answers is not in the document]
upils: I am very tempted to create a DiskImage class and use it to track the state of this kind of things. I suppose this need will become more precise as we keep adding features to the packing so we can decide how to refactor at this moment. I am fine with the current solution.

imagecraft/pack/gptutil.py
```diff
@@ -0,0 +1,256 @@
...
+
+    :param imagepath: Path to image file.
+    :raises CalledProcessError: If sfdisk fails.
+    """
+    # This is the end of the usable GPT sectors before the footer
+    # so we know every sector after this is the footer.
+    return int(_get_partition_table(imagepath)["lastlba"]) + 1
+
+
+def extract_backup_gpt(*, imagepath: Path, sector_size: int, footerpath: Path) -> None:
```
upils, line 193: It looks like this one and  `extract_primary_gpt` (and so  `backup_gpt_sector_start` and `primary_gpt_sector_count`) are not used. I think we can remove them.

upils (changes requested):
Thanks for the fixes! I just have a couple more questions and some cleaning but otherwise it looks good. We can fix the spread tests and the size reserved for the partition tables in other tasks.

snap/snapcraft.yaml
```diff
@@ -80,8 +80,11 @@ parts:
       - python3.12-minimal
       - python3-minimal
       - fuse-overlayfs
+      - mtools
+      - fdisk
     organize:
       "usr/bin/fuse-overlayfs": "libexec/imagecraft/fuse-overlayfs"
+      "usr/lib/python3/dist-packages/apt*": "lib/python3.12/site-packages/"
```
lengau, line 87:
```suggestion
      "usr/lib/python3/dist-packages/apt*": "lib/python3.12/site-packages/"
      "usr/sbin/sfdisk": "libexec/imagecraft/sfdisk"
```

snap/snapcraft.yaml
```diff
@@ -23,7 +23,7 @@ apps:
   imagecraft:
     command: bin/python3 $SNAP/bin/imagecraft
     environment:
-      PATH: "$SNAP/libexec/imagecraft:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+      PATH: "$SNAP/libexec/imagecraft:/snap/core24/bin:/snap/core24/sbin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
lengau, line 26:
```suggestion
      PATH: "$SNAP/libexec/imagecraft:/snap/core24/current/bin:/snap/core24/current/sbin:/snap/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```

tests/spread/pack/complex/imagecraft.yaml
```diff
@@ -0,0 +1,54 @@
...
+    stage-packages: [busybox-static]
+    organize:
+      "*": (volume/party/rootfs)/
+
+volumes:
+  party:
+    schema: gpt
+    structure:
+      - name: efi
+        type: C12A7328-F81F-11D2-BA4B-00A0C93EC93B
```
lengau, line 26: It might be nice to have convenience synonyms for the GUIDs for EFI system partitions and Linux partitions. (Out of scope here, more a design suggestion to think about separately)

tests/unit/pack/test_gptutil.py
```diff
@@ -0,0 +1,190 @@
...
+
+@pytest.fixture
+def volume():
+    return Volume(
+        schema="gpt",  # pyright: ignore[reportArgumentType]
+        structure=[  # pyright: ignore[reportArgumentType]
+            {
+                "name": "efi",
+                "role": "system-boot",
+                "type": "0FC63DAF-8483-4772-8E79-3D69D8477DE4",
```
lengau, line 31: Either this should probably be `C12A7328-F81F-11D2-BA4B-00A0C93EC93B` or we should have a comment noting that the type here is intentionally not the EFI system partition.

I'd imagine that later we might want to check the partition types during validation, and at that point this datum here will probably raise questions about whether it's intentionally invalid.

lengau (approved):
LGTM!

tests/unit/pack/test_gptutil.py
upils: Good point, I will fix it to avoid future confusion.

imagecraft/services/pack.py
```diff
@@ -43,15 +49,63 @@ def __init__(
...
+        # https://github.com/canonical/craft-parts/issues/665
+        project_dirs = self._services.lifecycle.project_info.dirs
+        with tempfile.TemporaryDirectory() as partition_dir:
+            for structure_item in volume.structure:
+                partition_name = get_partition_name(volume_name, structure_item)
+                partition_prime_dir = project_dirs.get_prime_dir(
+                    partition=partition_name
+                )
+                partition_img = (
+                    Path(partition_dir) / f"{volume_name}.{structure_item.name}.img"
```
cmatsuoka, line 84: Calling this `partition_dir` could be confusing, my first reaction was to think it was the top-level partitions work dir. I would rather leave it as `temp_dir` or something like that.

imagecraft/services/pack.py
```diff
@@ -43,15 +49,63 @@ def __init__(
...
+            imagepath=disk_image_file,
+            sector_size=SECTOR_SIZE,
+            layout=volume,
+        )
+
+        # Create partition images with filesystems.  These are always recreated, but we
+        # may want to revisit that once this is solved:
+        # https://github.com/canonical/craft-parts/issues/665
+        project_dirs = self._services.lifecycle.project_info.dirs
+        with tempfile.TemporaryDirectory() as partition_dir:
```
cmatsuoka, line 77: This could be a problem in case of very large images and limited `/tmp` capacity but we can address that later if needed.

cmatsuoka (changes requested):
It seems to work, but please change the sector alignment function to use integer math. Overall I would prefer a more object-oriented approach to image-partitions organization but this also works.

imagecraft/services/pack.py
upils: Good catch. This *will* be a problem even for moderate images since `/tmp` is now rather small by default.

cmatsuoka (approved):
Thanks!

PR #96, opened 2025-02-26
feat: support grammar in Volumes section

### LINKED ISSUE #82: Support craft grammar in Volumes section
### What needs to get done

Support the craft grammar in the volumes section to enable defining variations around the disk partitions defined.

### Why it needs to get done

For a given configuration (ex. ubuntu-minimal) the disk structure can vary between platforms (ex. amd64, arm64). This feature will avoid configuration duplication.

upils:

---

Support the craft grammar in the volumes section to enable defining variations around the disk partitions defined.

Fixes #82 
CRAFT-4061


imagecraft/grammar.py
```diff
@@ -0,0 +1,114 @@
...
+) -> dict[str, Any]:
+    """Process grammar for volumes.
+
+    :param yaml_data: unprocessed volumes section of imagecraft.yaml.
+    :returns: processed volumes section of imagecraft.yaml.
+    """
+
+    def self_check(value: Any) -> bool:  # noqa: ANN401
+        return bool(
+            value == value  # pylint: disable=comparison-with-itself  # noqa: PLR0124
```
lengau, line 48: I don't understand this in context. The only non-true value I can think of is NaN, but I'm not sure how we'd get that?

upils: Out of coherence I used the same implementation of `self_check` as in snapcraft and craft-application. I was very tempted to simply `return True` but I suspect there are corner cases I might have missed.

I doubt we could end up with a NaN value since that would be caught by the model validation, but I confirm that would return `False` (I tested with `a = float('nan'); a ==a`)

common.mk
```diff
@@ -48,11 +48,11 @@ help: ## Show this help.
 
 .PHONY: setup
 setup: install-uv setup-precommit ## Set up a development environment
-	uv sync $(UV_TEST_GROUPS) $(UV_LINT_GROUPS) $(UV_DOCS_GROUPS)
+	uv sync $(UV_TEST_GROUPS) $(UV_LINT_GROUPS) $(UV_DOCS_GROUPS) --reinstall-package craft-application
```
tigarmo, line 51: what's this for?

tigarmo (approved):

common.mk
lengau: Needed while we're using craft-application from a branch (so it reinstalls with the latest version from that branch)

lengau (approved):

common.mk
upils: Exactly. I will remove it before merging.

PR #109, opened 2025-03-18
docs: Clean and document new fields

upils:

---

- add documentation on overlay
- clean references to `ubuntu-image`
- document the `volumes` section of the `imagecraft.yaml` (and other common fields)
- add a better `imagecraft.yaml` example (will evolve in future PRs but is now closer to what imagecraft expects)


upils:
@medubelko is the integration of the parts properties now closer to what you had in mind? I was unable to hide the Part properties page from the navigation list though, so if you know a way let me know :).

docs/reference/commands/init.rst
```diff
@@ -0,0 +1,52 @@
...
+If '<project-dir>' is provided, initialise in that directory,
+otherwise initialise in the current working directory.
+
+If '--name <name>' is provided, the project will be named '<name>'.
+Otherwise, the project will be named after the directory it is initialised in.
+
+'--profile <profile>' is used to initialise the project for a specific use case.
+
+Init can work in an existing project directory. If there are any files in the
+directory that would be overwritten, then init command will fail.
```
medubelko, line 18: What do you think about my reviewing the command strings as a 25.10 task, for alignment with our CLI specs? I've done it for some of the other new apps.

docs/explanation/security-overview.rst
```diff
@@ -18,29 +18,5 @@ trusted.
...
-
-- These configuration files for secrets should then be securely stored, and if secrets
-  are used, they should ideally be injected at runtime.
-
-
-.. LINKS
-
-.. _ubuntu-image security overview: https://canonical-subiquity.readthedocs-hosted.com/en/latest/explanation/ubuntu-image-security-overview.html
+``imagecraft`` is a wrapper around several tools. The tool itself does not use
+cryptographic technologies.
```
medubelko, line 22: Double-checking: was this cleared with the relevant stakeholders?

docs/reference/commands/other-commands.rst
```diff
@@ -0,0 +1,5 @@
+:ref:`ref_commands_init`
+   Create an initial project filetree.
+
+:ref:`ref_commands_version`
+   Show the application version and exit.
```
medubelko, line 5: I don't quite understand why we needed to transclude these – couldn't we have included this content and `lifecycle-commands.rst` directly in `commands.rst`?

medubelko (commented):
> @medubelko is the integration of the parts properties now closer to what you had in mind?

Yes! But, it has some side effects I didn't anticipate. I'll come up with an alternative soon.

I left some questions and suggestions.

docs/explanation/overlay-step.rst
```diff
@@ -0,0 +1,20 @@
...
+
+Images are built in a sequence of five separate steps: pull, overlay,
+build, stage and prime.
+
+The overlay step is configured with overlay parameters.
+To learn more about pull, build, stage and prime see
+:doc:`/reference/part-lifecycle-details`
+
+The overlay step provides the means to refine the content of an image parts
+after parts. ``overlay-script`` will run the provided script in this step.
```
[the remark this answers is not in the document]
upils: In the context of imagecraft the overlay is not so much a "env to build stuff in" but the actual target of the building process. The idea is to bootstrap a rootfs in the first part and then refine/add content/packages in it via the following parts. The part themselves will just hold what was added/modified but the consolidated result (when the overlay is fully mounted/merged) will be a full rootfs.

If you want I can show you a demo to clarify that.

docs/explanation/overlay-step.rst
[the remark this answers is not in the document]
upils: I do not have strong opinion on that but this page includes another one from the explanation section of craft-parts. So out of coherence I am tempted to keep it here. What do you think?

docs/explanation/security-overview.rst
upils: Imagecraft is not using `ubuntu-image` anymore. The `ubuntu-bootstrap` is also gone. Even if imagecraft's code is not doing any cryptographic operation, the craft framework probably do. I did not fully evaluated this, so what do you think of completely removing this `Cryptographic` section until I can properly assert that next cycle?

docs/reference/commands/other-commands.rst
upils: I do not know enough sphinx best practices to know if this is needed or not. I suggest reworking that for both files if we want in another PR.

docs/reference/commands/init.rst
upils: This piece of doc is automatically generated from the CLI and in this case is not specific to imagecraft.
So I guess any CLI rework will happen in craft-application. Unless you meant something else?

docs/explanation/overlay-step.rst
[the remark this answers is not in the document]
medubelko: I believe I understand. My comment was more pointed at the grammar. As-is it's not clear. :)

From your description, it sounds like the sentence should read more like this:

> The overlay step in each part provides the means to refine the contents of the image.

What do you think?

docs/reference/commands/init.rst
medubelko: That's what I meant. :) It's older content that preades both of us.

The reason I suggested it is because we're bubbling up content that normally doesn't get seen outside of the CLI, so we can double our value if I review it once and improve it in both docs and CLI.

docs/reference/commands/init.rst
medubelko, line 1: Our discussion reminds me of something else. I should have picked up on this earlier.

Pages created by the `gen_cli` script shouldn't be committed. The other apps have placed their autogenerated paths in [.gitignore](https://github.com/canonical/rockcraft/blob/adb3031276a7f2f4039c35cf2388254aaa88861f/.gitignore#L151).

docs/explanation/security-overview.rst
medubelko: Agreed, let's remove the section for now. I'm in discussion with Sergio about a good template we could use for all libraries that then gets transcluded in apps as-needed.

docs/reference/commands/other-commands.rst
medubelko: Sure, let's earmark this.

docs/explanation/overlay-step.rst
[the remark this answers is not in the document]
medubelko: I see, that's a good point. Let's leave it as-is for now, and we'll rethink Craft Parts' making it an explanation later.

medubelko:
>  I'll come up with an alternative soon.

I gave this some more thought, and restored the page. Snapcraft has this page too, but is missing its project file reference.

All apps will have their project file refernces served by the upcoming Pydantic model extension, so we'll have a solution soon.

docs/reference/commands/init.rst
upils: I missed that, thanks. See https://github.com/canonical/imagecraft/pull/109/commits/8168665738ec1b96c955cf5b837e4617b93161b1

medubelko (approved):
@upils I tweaked the project file keys to be more in line with what the extension will produce.

docs/explanation/parts.rst
```diff
@@ -5,7 +5,3 @@
 ..
   exclude the overlay step explanation from craft-parts
```
tigarmo, line 6: (minor) this comment is outdated now

docs/reference/imagecraft-yaml-file.rst
```diff
@@ -27,45 +25,80 @@ The name of the image. This value must:
...
+**Required**: Yes
+
+The system and version that will be used during the build, but not
+included in the final image itself. It comprises the set of tools and libraries
+that Imagecraft will use when building the image contents.
+
+.. note::
+   ``devel`` is a "special" value that means "the next Ubuntu version, currently
+   in development". This means that the contents of this system changes
+   frequently and should not be relied on for production rocks.
```
tigarmo, line 87:
```suggestion
   frequently and should not be relied on for production images.
```

PR #113, opened 2025-03-24
Implement grub installation in the image

upils:

---

Implement grub installation in the image after the disk is created.  

The chroot handling is forked from craft-parts.

### DIFF
--- a/imagecraft/pack/chroot.py
+++ b/imagecraft/pack/chroot.py
@@ -22,7 +22,6 @@ from multiprocessing.connection import Connection
 from pathlib import Path
 from typing import Any
 
-from craft_cli import emit
 from craft_parts.utils import os_utils
 
 from imagecraft import errors
@@ -70,8 +69,6 @@ class Mount:
         self._mountpoint = base_path / self._relative_mountpoint.lstrip("/")
         pid = os.getpid()
         # Only mount if mountpoint exists.
-        emit.debug(f"Mount {self._src} on {str(self._mountpoint)}")
-        logger.debug("[pid=%d] mount %r on chroot", pid, str(self._mountpoint))
         if self._mountpoint.exists():
             logger.debug("[pid=%d] mount %r on chroot", pid, str(self._mountpoint))
             os_utils.mount(self._src, str(self._mountpoint), *args)
@@ -110,7 +107,6 @@ def _runner(
     logger.debug("[pid=%d] child process: target=%r", pid, target)
     try:
         logger.debug("[pid=%d] chroot to %r", pid, path)
-        os.chdir(path)
         os.chroot(path)
         res = target(*args, **kwargs)
     except Exception as exc:  # noqa: BLE001
@@ -139,12 +135,6 @@ class Chroot:
         # There's no need to restore the file to its original condition because
         # this operation happens on a temporary filesystem layer.
         logger.debug("setup chroot: %r", self.path)
-        resolv_conf = self.path / "etc/resolv.conf"
-        if resolv_conf.is_symlink():
-            resolv_conf.unlink()
-            resolv_conf.touch()
-        elif not resolv_conf.exists() and resolv_conf.parent.is_dir():
-            resolv_conf.touch()
 
         for entry in self.mounts:
             entry.mount(base_path=self.path)
--- a/imagecraft/pack/chroot.py
+++ b/imagecraft/pack/chroot.py
@@ -1,17 +1,15 @@
-# -*- Mode:Python; indent-tabs-mode:nil; tab-width:4 -*-
+# Copyright 2025 Canonical Ltd.
 #
-# Copyright 2021 Canonical Ltd.
-#
-# This program is free software; you can redistribute it and/or
-# modify it under the terms of the GNU Lesser General Public
-# License version 3 as published by the Free Software Foundation.
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
 #
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-# Lesser General Public License for more details.
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
 #
-# You should have received a copy of the GNU Lesser General Public License
+# You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 """Execute a callable in a chroot environment."""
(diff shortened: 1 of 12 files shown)


imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+    mounts: list[Mount] = [
+        Mount(
+            fstype=None,
+            src=f"{loop_dev}p{rootfs_partition_num}",
+            relative_mountpoint="/",
+        ),
+        Mount(
+            fstype=None,
+            src=f"{loop_dev}p{boot_partition_num}",
+            relative_mountpoint="/boot/efi",
```
dariuszd21, line 107: I don't know if we want to make it configurable at some point, but it's worth thinking about it

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+        "--rename",
+        os_prober,
+    ]
+
+    divert_os_prober_command = [divert_base_command, *list(divert_common_args)]
+
+    undivert_os_prober_command = [
+        divert_base_command,
+        "--remove",
+        *list(divert_common_args),
```
dariuszd21, line 65: It should work without list()
```suggestion
        *divert_common_args,
```

imagecraft/pack/image.py
```diff
@@ -0,0 +1,110 @@
...
+        *,
+        volume: Volume,
+        disk_path: Path,
+    ) -> None:
+        self.volume = volume
+        self.disk_path = disk_path
+        if not self.disk_path.is_file():
+            raise errors.ImageError(
+                f"Image file specified ({self.disk_path}) doesn't exist."
+            )
```
dariuszd21, line 64: It may also exist and be a `directory`, so this error may be a bit misleading

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        run(*grub_install_command)
+        run(*divert_os_prober_command)
+        run(*update_grub_command)
+        run(*undivert_os_prober_command)
+    except subprocess.CalledProcessError as err:
+        raise errors.GRUBInstallError("Cannot install grub") from err
+
+
+def setup_grub(image: Image, workdir: Path, arch: str) -> None:
```
dariuszd21, line 77: The thing here and also an image looks like items that might be in the future transformed to a service (maybe BootloaderService with `.setup()` API). 
Don't worry about it now, but it might be something to consider for the future

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+from imagecraft import errors
+from imagecraft.pack.chroot import Chroot, Mount
+from imagecraft.pack.image import Image
+from imagecraft.subprocesses import run
+
+_ARCH_TO_GRUB_EFI_TARGET: dict[str, str] = {
+    "amd64": "x86_64-efi",
+    "arm64": "arm64-efi",
+    "armhf": "arm-efi",
+}
```
dariuszd21, line 31: I think you should use `craft_platforms.DebianArchitecture` as keys here

dariuszd21 (commented):
Some of my initial thoughts, but in general it looks pretty good

imagecraft/pack/chroot.py
```diff
@@ -0,0 +1,209 @@
...
+        :param args: Arguments for target.
+        :param kwargs: Keyword arguments for target.
+
+        :returns: The target function return value.
+        """
+        logger.debug("[pid=%d] parent process", os.getpid())
+        parent_conn, child_conn = multiprocessing.Pipe()
+        child = multiprocessing.Process(
+            target=_runner, args=(self.path, child_conn, target, args, kwargs)
+        )
```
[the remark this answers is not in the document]
dariuszd21: You've answered that one live, that this is `craft-parts` inspired.

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,120 @@
...
+        raise errors.GRUBInstallError("Cannot install grub") from err
+
+
+def setup_grub(image: Image, workdir: Path, arch: str) -> None:
+    """Setups grub in the image."""
+    rootfs_partition_num = image.data_partition_number
+    boot_partition_num = image.boot_partition_number
+
+    if boot_partition_num is None:
+        emit.debug("Cannot install grub without a boot partition")
```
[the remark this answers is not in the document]
upils: This is on purpose. Since there is no way to express that grub should *not* be installed we decided to try setting it up if conditions are met and just return if not. This way users can already try to experiment with non-grub images.

imagecraft/pack/chroot.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+class Chroot:
+    """Chroot manager."""
+
+    mounts: list[Mount]
+    path: Path
+
+    def __init__(
+        self, *, path: Path, mounts: list[Mount] = default_linux_mounts
```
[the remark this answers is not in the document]
upils: I think I will simply remove `default_linux_mounts`.

imagecraft/pack/grubutil.py
upils: yeah this implementation is based on some ubuntu-specific assumptions. Striking the right balance between a robust and a configurable tool is challenging.

imagecraft/pack/image.py
```diff
@@ -0,0 +1,110 @@
...
+
+from imagecraft import errors
+from imagecraft.models import Role, Volume
+from imagecraft.subprocesses import run
+
+
+def _get_loop_devices() -> list[dict[str, Any]]:
+    return cast(
+        list[dict[str, Any]],
+        json.loads(run("losetup", "--json").stdout.strip())["loopdevices"],
```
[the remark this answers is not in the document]
upils: Good point, but why not a constant string?

imagecraft/pack/image.py
```diff
@@ -0,0 +1,110 @@
...
+            )
+        return self.loop_device
+
+    def get_loopdevs(self) -> Generator[dict[str, Any]]:
+        """Return the loop devices attached from this image file."""
+        for loop_device in _get_loop_devices():
+            try:
+                if self.disk_path.samefile(Path(loop_device["back-file"])):
+                    yield loop_device
+            except FileNotFoundError:  # noqa: PERF203
```
tigarmo, line 104: you can use a `with(contextlib.supress(FileNotFoundError))` to drop the noqa

tigarmo (approved):

imagecraft/pack/chroot.py
```diff
@@ -0,0 +1,189 @@
...
+
+        self._mountpoint = base_path / self._relative_mountpoint.lstrip("/")
+        pid = os.getpid()
+        # Only mount if mountpoint exists.
+        emit.debug(f"Mount {self._src} on {str(self._mountpoint)}")
+        logger.debug("[pid=%d] mount %r on chroot", pid, str(self._mountpoint))
+        if self._mountpoint.exists():
+            logger.debug("[pid=%d] mount %r on chroot", pid, str(self._mountpoint))
+            os_utils.mount(self._src, str(self._mountpoint), *args)
+        else:
```
flotter, line 78: If someone asks for a mount operation, and its not done here, will that not cause a failure further down the line? Is this not an error condition that is better caught early ?

imagecraft/pack/chroot.py
```diff
@@ -0,0 +1,189 @@
...
+    target: Callable[..., str | None],
+    args: tuple[str],
+    kwargs: dict[str, Any],
+) -> None:
+    """Chroot to the execution directory and call the target function."""
+    pid = os.getpid()
+    logger.debug("[pid=%d] child process: target=%r", pid, target)
+    try:
+        logger.debug("[pid=%d] chroot to %r", pid, path)
+        os.chdir(path)
```
flotter, line 113: I am not sure what the significance of changing the directory first is ?

imagecraft/pack/chroot.py
```diff
@@ -0,0 +1,189 @@
...
+        # a regular file to bind-mount the host resolver configuration on.
+        #
+        # There's no need to restore the file to its original condition because
+        # this operation happens on a temporary filesystem layer.
+        logger.debug("setup chroot: %r", self.path)
+        resolv_conf = self.path / "etc/resolv.conf"
+        if resolv_conf.is_symlink():
+            resolv_conf.unlink()
+            resolv_conf.touch()
+        elif not resolv_conf.exists() and resolv_conf.parent.is_dir():
```
flotter, line 146: It looks like we are creating a file if the parent directory exists. What should happen if it does not exist ?

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,156 @@
...
+        divert_base_command,
+        "--remove",
+        *divert_common_args,
+    ]
+
+    # Check if grub-install is available, otherwise skip the installation without error
+    try:
+        run(*check_grub_install)
+    except FileNotFoundError:
+        emit.message("Skipping GRUB installation because grub-install is not available")
```
flotter, line 80: I would have thought that if we reach this point, the image creation is doomed. The error sounds like this is an optional step that can be skipped ?

imagecraft/pack/image.py
flotter: How about creating it on /dev/sdc or /dev/mmcblk0 ? Maybe later :)

flotter (approved):
Looks amazing! Left a couple of questions, mostly for my own understanding. Feel free to live-run me through them to save time.

imagecraft/pack/grubutil.py
upils: We only want to try to install GRUB if some conditions are met, otherwise we conclude that the user is trying to build another kind of image (maybe with bootloader provided in another way) so it is fine to do nothing.
This piece of code is the last check, after that a failure will really fail the build.

imagecraft/pack/chroot.py
upils: This workaround was copied from the original implementation in craft-parts I forked. Since we are now mounting `/run` the name resolution works fine without tweaking this file, so I removed it.

imagecraft/pack/chroot.py
upils: It was copied from the original implementation but after testing I cannot determine why that is necessary so I removed it.

imagecraft/services/pack.py
```diff
@@ -46,6 +46,9 @@ def pack(self, prime_dir: Path, dest: Path) -> list[Path]:  # noqa: ARG002
         volume_name, volume = next(iter(project.volumes.items()))
         disk_image_file = dest / (volume_name + os.extsep + "img")
 
+        # Remove previously built image
+        disk_image_file.unlink(missing_ok=True)
+
```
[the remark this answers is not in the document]
upils: This is due to the way the file is created, using `fallocate`. But now I think that would be better if the `create_zero_image` actually creating the file (calling `fallocate`) would also be in charge of properly removing a pre-existing one. I will do that.

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+        divert_base_command,
+        "--remove",
+        *list(divert_common_args),
+    ]
+
+    try:
+        run(*grub_install_command)
+        run(*divert_os_prober_command)
+        run(*update_grub_command)
+        run(*undivert_os_prober_command)
```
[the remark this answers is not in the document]
upils: Discussed directly. Let's keep treating this whole block as a single operation for users for now.

imagecraft/pack/grubutil.py
```diff
@@ -0,0 +1,133 @@
...
+    chroot = Chroot(path=mount_dir, mounts=mounts)
+
+    try:
+        chroot.chroot(
+            target=_grub_install,
+            grub_target=_ARCH_TO_GRUB_EFI_TARGET[arch],
+            loop_used=loop_dev,
+        )
+    finally:
+        image.detach_loopdevs()
```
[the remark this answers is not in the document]
upils: Nice suggestion! I created an issue for that because I would need to explore how to implement a context manager.
See https://github.com/canonical/imagecraft/issues/117

imagecraft/pack/chroot.py
upils: Good point. And also the umounting should probably not stop at the first umounting failure but try to umount as many mountpoints as possible before raising an exception if at least one could not be umounted. I created an [issue](https://github.com/canonical/imagecraft/issues/118) for that to deal with that in another PR.

PR #124, opened 2025-04-01
fix: respect constraints on unit of structure size from original spec

upils:

---

The Volumes section is a subset of the original Volumes specification. This must be respected strictly.
Keep using the ByteSize type but pre-validate the received value to enforce the stricter set of valid units and to convert it to bytes interpreted by the ByteSize type.

### DIFF
--- a/imagecraft/models/volume.py
+++ b/imagecraft/models/volume.py
@@ -40,6 +40,9 @@ from pydantic import (
     model_validator,
 )
 
+MIB = 1 << 20  # 1 MiB (2^20)
+GIB = 1 << 30  # 1 GiB (2^30)
+
 # Avoid matches on substrings when validating Volume/Structure names.
 PARTITION_COMPILED_STRICT_REGEX = re.compile(
     r"^" + VALID_PARTITION_REGEX.pattern + r"$", re.ASCII
@@ -69,10 +72,12 @@ StructureName = typing.Annotated[
 def _validate_structure_size(value: str) -> str:
     """Validate and convert the structure size.
 
-    The Volume specification does not respect the IEC 80000-13 Standard:
-    M means MiB (2^20) in the Volume spec. (means 10^2 in the standard).
-    So this validation converts the unit before the conversion to the ByteSize
-    type (respecting the IEC 80000-13 Standard).
+    The Volumes specification supports the following values for size:
+    - <bytes>
+    - <bytes/2^20>M
+    - <bytes/2^30>G
+
+    Validate the unit and convert it to a multiplier applied to the numerical value.
     """
     value = str(value)
     match = STRUCTURE_SIZE_COMPILED_REGEX.match(value)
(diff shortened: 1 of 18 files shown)


tests/unit/models/test_volume.py
```diff
@@ -50,10 +50,10 @@ def test_volume_valid():
     )
     assert volume.volume_schema == "gpt"
     assert len(volume.structure) == 3
-    assert volume.structure[0].size == 6442450944
+    assert volume.structure[0].size == 3221225472
```
lengau, line 53: It's not particularly obvious where this number came from. What about:

```suggestion
    assert volume.structure[0].size == 3 * 1024**3
```

(Similar for below)

lengau (approved):
LGTM

tests/unit/pack/test_gptutil.py
```diff
@@ -120,7 +120,7 @@ def test_create_empty_gpt_image(mocker, volume, tmp_path):
     )
     create_zero_image.assert_called_with(
         imagepath=imagepath,
-        disk_size=diskutil.DiskSize(bytesize=6463517184, sector_size=512),
+        disk_size=diskutil.DiskSize(bytesize=6464488448, sector_size=512),
```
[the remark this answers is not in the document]
upils: I did the maths anyway and added a comment to explain it so both can be checked if things change. Luckily in this example values are already aligned on the sector size so the maths is rather simple to explain.

imagecraft/models/volume.py
```diff
@@ -55,13 +56,47 @@
...
     StringConstraints(
         max_length=GPT_NAME_MAX_LENGTH, pattern=STRUCTURE_NAME_COMPILED_REGEX
     ),
 ]
 
+
+def _validate_structure_size(value: str) -> str:
+    """Validate and convert the structure size.
+
+    The Volume specification does not respect the IEC 80000-13 Standard:
```
cmatsuoka, line 72: Maybe use different wording here? Instead of mentioning a standard we don't follow, we could just say what the expected format is, and let the validator do its magic to produce the final value. Personally I would just resolve the value directly instead of converting units.

imagecraft/models/volume.py
```diff
@@ -55,13 +56,47 @@
...
+    if not match:
+        raise ValueError(SIZE_INVALID_MSG)
+
+    # convert M/G to MiB/GiB
+    unit: str = ""
+    unit_matched = match.group("unit")
+    if unit_matched == "M":
+        unit = "MiB"
+    elif unit_matched == "G":
+        unit = "GiB"
```
cmatsuoka, line 89: Personally I would just declare the field as numeric and resolve the value directly instead of translating this to something that pydantic understands to then let it convert values. Doing it ourselves is simpler, faster, and would yield the same result (we may already have the unit constants if we do size computation elsewhere in the code).

imagecraft/models/volume.py
upils: Done. Let me know what you think.

cmatsuoka (approved):
Thanks

Issue #117, opened 2025-03-28
Improve image loop dev detaching handling


upils:
It may be worth wrapping calls to `chroot.chroot()` with a context manager to make it more "pythonic". Something similar to `closing()`, so there is no need to remember to do `finally:` block everywhere.
It could look similar to something like this:
```suggestion
    with image.detaching():
        chroot.chroot(
            target=_grub_install,
            grub_target=_ARCH_TO_GRUB_EFI_TARGET[arch],
            loop_used=loop_dev,
        )
```

_Originally posted by @dariuszd21 in https://github.com/canonical/imagecraft/pull/113#discussion_r2012416760_

PR #126, opened 2025-04-02
Be stricter when mounting and safer when umounting

### LINKED ISSUE #118: Be stricter when mounting and safer when umounting
### What needs to get done

- Raise an exception when the mountpoint does not exist in `Mount.mount()`
- Catch exceptions when calling `Mount.umount()` in `_cleanup_chroot` to umount as many dirs as possible from the given list.

### Why it needs to get done

If the mountpoint given to `Mount` does not exist, nothing is mounted. This may later break the command executed in the chroot. Detecting early that a target cannot be mounted should be easier to manage.

When  cleaning up the chroot, mounted dirs are unmounted in reverse order. A umounting failure is not caught, so if one umounting fails the current implementation does not try to umount the rest of the list. We should try to left the system in the most clean state so a umounting exception should be caught and raised only after attempting to umount every other mounted dirs.

upils:

---

- Raise an exception when the mountpoint does not exist in Mount.mount()
- Catch exceptions when calling Mount.umount() in `_cleanup` to umount as many dirs as possible from the given list.

Fixes #118 
CRAFT-4402

### DIFF
--- a/imagecraft/pack/chroot.py
+++ b/imagecraft/pack/chroot.py
@@ -127,14 +127,8 @@ class Chroot:
         self.path = path
         self.mounts = mounts
 
-    def _setup_chroot(self) -> None:
+    def _setup(self) -> None:
         """Chroot environment preparation."""
-        # Some images (such as cloudimgs) symlink ``/etc/resolv.conf`` to
-        # ``/run/systemd/resolve/stub-resolv.conf``. We want resolv.conf to be
-        # a regular file to bind-mount the host resolver configuration on.
-        #
-        # There's no need to restore the file to its original condition because
-        # this operation happens on a temporary filesystem layer.
         logger.debug("setup chroot: %r", self.path)
 
         for entry in self.mounts:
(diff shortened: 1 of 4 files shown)


imagecraft/pack/chroot.py
```diff
@@ -126,28 +124,34 @@ def __init__(self, *, path: Path, mounts: list[Mount]) -> None:
...
+                    msg += f" ({err.stderr.strip()!s})"
+                umount_errors.append(msg)
+
+        if umount_errors:
+            raise errors.ChrootExecutionError(
+                message="Failed to clean chroot", details="\n".join(umount_errors)
+            )
 
-    def chroot(
+    def execute(
```
lengau, line 154: It's very out-of-scope here, but having a context manager for the chroot and having this method wrap it could be useful.

lengau (approved):
LGTM!

imagecraft/pack/chroot.py
upils: Well.... https://github.com/canonical/imagecraft/issues/117 :D

imagecraft/pack/chroot.py
```diff
@@ -107,6 +104,7 @@ def _runner(
     logger.debug("[pid=%d] child process: target=%r", pid, target)
     try:
         logger.debug("[pid=%d] chroot to %r", pid, path)
+        os.chdir(path)
```
cmatsuoka, line 107: Why are we changing to the directory before the chroot? Maybe we should do it after the chroot (changing to `/`) to ensure cwd is significant after the change?

cmatsuoka (approved):
Changes look good, but please consider re-evaluating the need for `chdir` before `chroot` and placing `chdir("/")` after `chroot` to set a valid CWD in the caller process.

cmatsuoka:
I stand corrected:
```
>>> import os
>>> os.chdir("/tmp")
>>> os.chroot("/tmp")
>>> os.getcwd()
'/'
```

PR #131, opened 2025-04-10
Fix grub install

upils:

---

When trying to install GRUB, if `/boot/efi` mountpoint is missing the installation must be skipped (because it means the conditions to install GRUB are not met) but the packing must not fail.

### DIFF
--- a/imagecraft/pack/grubutil.py
+++ b/imagecraft/pack/grubutil.py
@@ -109,7 +109,7 @@ def setup_grub(image: Image, workdir: Path, arch: str) -> None:
         emit.message("Skipping GRUB installation because no boot partition was found")
         return
     if rootfs_partition_num is None:
-        emit.message("Skipping GRUB installation because data partition was found")
+        emit.message("Skipping GRUB installation because no data partition was found")
         return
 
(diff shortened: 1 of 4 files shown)


imagecraft/pack/grubutil.py
```diff
@@ -155,5 +155,9 @@ def setup_grub(image: Image, workdir: Path, arch: str) -> None:
             grub_target=_ARCH_TO_GRUB_EFI_TARGET[arch],
             loop_dev=loop_dev,
         )
+    except errors.ChrootMountError as err:
+        # Ignore mounting errors indicating the rootfs does not have
+        # the needed structure to install grub.
+        emit.message(f"Cannot install GRUB on this rootfs: {err}")
```
mr-cal, line 161: pedantic: these probably should be `emit.progress(..., permanent=True)` since they're warnings and not the final output (https://github.com/canonical/craft-cli/issues/248)

mr-cal (approved):

imagecraft/pack/grubutil.py
upils: Oh nice catch, thanks!

tigarmo (approved):

PR #135, opened 2025-04-15
feat: context manager for image loop dev management

upils:

---

### DIFF
--- a/imagecraft/pack/image.py
+++ b/imagecraft/pack/image.py
@@ -84,7 +84,8 @@ class Image:
                 return i + 1
         return None
 
-    def attach_loopdev(self) -> str:
+    @contextlib.contextmanager
+    def attach_loopdev(self) -> Generator[str]:
         """Attach a loop device for this image file."""
         if not hasattr(self, "loop_device"):
             # This command attaches a loop device and returns the path in /dev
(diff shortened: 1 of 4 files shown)


dariuszd21 (approved):
Looks good to me thanks!

imagecraft/pack/image.py
```diff
@@ -84,7 +84,8 @@ def boot_partition_number(self) -> int | None:
                 return i + 1
         return None
 
-    def attach_loopdev(self) -> str:
+    @contextlib.contextmanager
+    def attach_loopdev(self) -> Generator[str]:
```
cmatsuoka, line 88: I think we can return an `Iterator[str]` here instead, since we're yielding just to create the context manager.

imagecraft/pack/image.py
```diff
@@ -98,16 +99,19 @@ def attach_loopdev(self) -> str:
...
                 f"Attached image {self.disk_path} as loop device {self.loop_device}"
             )
-        return self.loop_device
+        try:
+            yield self.loop_device
+        finally:
+            self._detach_loopdevs()
 
-    def get_loopdevs(self) -> Generator[dict[str, Any]]:
+    def _get_loopdevs(self) -> Generator[dict[str, Any]]:
```
cmatsuoka, line 107: And this is a true generator.

imagecraft/pack/image.py
```diff
@@ -98,16 +99,19 @@ def attach_loopdev(self) -> str:
             emit.debug(
                 f"Attached image {self.disk_path} as loop device {self.loop_device}"
             )
-        return self.loop_device
+        try:
+            yield self.loop_device
+        finally:
+            self._detach_loopdevs()
```
cmatsuoka, line 105: Why is the context manager function `attach_loopdev()` (singular) but it ends with `_detach_loopdevs()` (plural)? I attach a device and then remove all attached devices? Shouldn't we detach only the one that was attached?

imagecraft/pack/image.py
upils: Good catch. I looked into it because I was not aware of the difference.

imagecraft/pack/image.py
upils: We do detach **only** loop devices associated to the given "back-file". Even though we only create a single loop device and attach the file to it, there is no way to guaranty the file has not been attached a second time by another process. So with the current approach listing loop devices associated to the file could yield multiple results. The current approach is a rather safe one because it makes sure to detach everything to avoid corrupting the file. 
Another approach would be to use `self.loop_device` to only detach this specific loop device.

cmatsuoka (approved):
Thanks for the clarifications.

PR #127, opened 2025-04-04
Fix spread tests

upils:
- [x] Have you followed the guidelines for contributing?
- [x] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [ ] Have you successfully run `make lint && make test`?

---

PR #137, opened 2025-04-15
Clean output

upils:

---

Improve and harmonize logs and displayed messages at various verbosity levels.

### DIFF
--- a/imagecraft/application.py
+++ b/imagecraft/application.py
@@ -16,7 +16,12 @@
 
 """Main Imagecraft Application."""
 
+from typing import Any
+
+import craft_application
+import craft_cli
 from craft_application import Application, AppMetadata
+from craft_application.services import service_factory
 from craft_parts.plugins.plugins import PluginType
 from overrides import override  # type: ignore[reportUnknownVariableType]
 
@@ -32,6 +37,16 @@ APP_METADATA = AppMetadata(
 
 class Imagecraft(Application):
     """Imagecraft application definition."""
+
+    def __init__(
+        self,
+        app: craft_application.AppMetadata,
+        services: service_factory.ServiceFactory,
+    ) -> None:
+        super().__init__(app=app, services=services, extra_loggers={"imagecraft"})
+        self._global_args: dict[str, Any] = {}
+        self._dispatcher: craft_cli.Dispatcher | None = None
+        self._cli_loggers |= {"imagecraft"}
 
     @override
     def _get_app_plugins(self) -> dict[str, PluginType]:
--- a/imagecraft/application.py
+++ b/imagecraft/application.py
@@ -16,12 +16,7 @@
 
 """Main Imagecraft Application."""
 
-from typing import Any
-
-import craft_application
-import craft_cli
 from craft_application import Application, AppMetadata
-from craft_application.services import service_factory
 from craft_parts.plugins.plugins import PluginType
 from overrides import override  # type: ignore[reportUnknownVariableType]
 
@@ -37,16 +32,6 @@ APP_METADATA = AppMetadata(
 
 class Imagecraft(Application):
     """Imagecraft application definition."""
-
-    def __init__(
-        self,
-        app: craft_application.AppMetadata,
-        services: service_factory.ServiceFactory,
-    ) -> None:
-        super().__init__(app=app, services=services, extra_loggers={"imagecraft"})
-        self._global_args: dict[str, Any] = {}
-        self._dispatcher: craft_cli.Dispatcher | None = None
-        self._cli_loggers |= {"imagecraft"}
 
     @override
     def _get_app_plugins(self) -> dict[str, PluginType]:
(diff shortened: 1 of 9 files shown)


imagecraft/pack/gptutil.py
```diff
@@ -106,7 +106,8 @@ def _create_gpt_layout(
         start += sectors
     stdin_lines: str = "\n".join(_create_sfdisk_lines(header, partitions))
 
-    emit.debug(f"Stdin (sfdisk):\n{stdin_lines}")
+    emit.trace(f"Stdin for sfdisk:\n{stdin_lines}")
+    emit.progress("Partition the image")
```
mr-cal, line 110: Present tense for progress?
```suggestion
    emit.progress("Partitioning the image")
```

imagecraft/pack/grubutil.py
```diff
@@ -105,15 +110,23 @@ def setup_grub(image: Image, workdir: Path, arch: str) -> None:
     rootfs_partition_num = image.data_partition_number
     boot_partition_num = image.boot_partition_number
 
+    emit.progress("Setup GRUB in the image")
```
mr-cal, line 113:
```suggestion
    emit.progress("Setting up GRUB in the image")
```

imagecraft/services/pack.py
```diff
@@ -60,6 +61,7 @@ def pack(self, prime_dir: Path, dest: Path) -> list[Path]:  # noqa: ARG002
         with tempfile.TemporaryDirectory() as tmp_dir:
             for structure_item in volume.structure:
                 partition_name = get_partition_name(volume_name, structure_item)
+                emit.verbose(f"Prepare partition {partition_name}")
```
mr-cal, line 64:
```suggestion
                emit.verbose(f"Preparing partition {partition_name}")
```

imagecraft/services/pack.py
```diff
@@ -78,6 +80,7 @@ def pack(self, prime_dir: Path, dest: Path) -> list[Path]:  # noqa: ARG002
                     disk_size=partition_size,
                     label=structure_item.filesystem_label,
                 )
+                emit.verbose(f"Add partition {partition_name} to the image")
```
mr-cal, line 83:
```suggestion
                emit.verbose(f"Adding partition {partition_name} to the image")
```

mr-cal (approved):

imagecraft/application.py
```diff
@@ -33,6 +38,16 @@
 class Imagecraft(Application):
     """Imagecraft application definition."""
 
+    def __init__(
+        self,
+        app: craft_application.AppMetadata,
+        services: service_factory.ServiceFactory,
+    ) -> None:
+        super().__init__(app=app, services=services, extra_loggers={"imagecraft"})
```
lengau, line 46: The preferred way to do this is to pass `extra_loggers={"imagecraft"}` when instantiating the app in `cli._create_app`.

imagecraft/application.py
```diff
@@ -33,6 +38,16 @@
...
+        self._global_args: dict[str, Any] = {}
```
lengau, line 47: Other than the type, what's the difference between this and `self._global_arguments`? Are these the args passed by the dispatcher? If so, when will the Application need to access them?

imagecraft/pack/diskutil.py
```diff
@@ -254,5 +255,6 @@ def inject_partition_into_image(
         f"of={str(imagepath)}",
         f"bs={disk_size.sector_size}",
         f"seek={sector_offset}",
+        "status=none",
```
lengau, line 258: Should the `run` function make an `emit.trace` call about what it's running? Just thinking that this way we'll get better insight if one of these commands takes a long time.

lengau (changes requested):

imagecraft/application.py
upils: Thanks! I think I applied what I found in charmcraft without properly checking if there was a better way.

imagecraft/application.py
upils: Good catch. See my other comments above. This is now removed.

imagecraft/pack/diskutil.py
upils: The `run` function was previously using emit but when used as the callable for the chroot it was hanging (at verbosity levels above `debug`) (see https://github.com/canonical/imagecraft/pull/127/commits/51a9d979f0e97d996d766f18dab0deb6a9a14e1f#diff-c4c127f022fd38e4a8535ac927b9d4809e747ee1943fae4a13bb9203ab663bd7L44). At the time I wanted to fix it because it was basically breaking imagecraft if not used bit the proper verbosity level but we can revisit that. 

For now I am keeping calls to emit out of this function.

I reworked the call to `dd`, let me know what you think.

lengau: Thanks for the explanation. I'm fine with that!

lengau (approved):

PR #158, opened 2025-06-09
build(docs): add sphinx sitemap

jahn-junior:
- [X] Have you followed the guidelines for contributing?
- [X] Have you signed the [CLA](http://www.ubuntu.com/legal/contributors/)?
- [X] Have you successfully run `make lint && make test`?

---

IMAGECRAFT-26

* Adds `sphinx-sitemap` as a docs dependency
* Modifies `conf.py` to generate a site map during build
* Switches local builds to the `dirhtml` builder

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -53,10 +53,20 @@ html_theme_options = {
     "source_edit_link": "https://github.com/canonical/imagecraft",
 }
 
+# Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
+html_baseurl = "https://canonical-imagecraft.readthedocs-hosted.com/"
+
+if "READTHEDOCS_VERSION" in os.environ:
+    version = os.environ["READTHEDOCS_VERSION"]
+    sitemap_url_scheme = "{version}{link}"
+else:
+    sitemap_url_scheme = "latest/{link}"
+
 extensions = [
     "canonical_sphinx",
-    "pydantic_kitbash",
     "notfound.extension",
+    "pydantic_kitbash",
+    "sphinx_sitemap",
 ]
 # endregion
 
@@ -163,7 +173,7 @@ github_repository = "imagecraft"
 
 
 def generate_cli_docs(nil):
-    gen_cli_docs_path = (project_dir / "tools" / "docs" / "gen_cli_docs.py").resolve()
+    gen_cli_docs_path = (project_dir / "tools/docs/gen_cli_docs.py").resolve()
     os.system("%s %s" % (gen_cli_docs_path, project_dir / "docs"))
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ docs = [
     "pydantic-kitbash==0.0.5",
     "sphinx-autobuild~=2024.2",
     "sphinx-pydantic==0.1.1",
+    "sphinx-sitemap==2.6.0",
     "sphinx-toolbox~=4.0",
     "sphinx-lint==1.0.0",
     "sphinxcontrib-details-directive",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,7 @@ docs = [
     "pydantic-kitbash==0.0.5",
     "sphinx-autobuild~=2024.2",
     "sphinx-pydantic==0.1.1",
-    "sphinx-sitemap==2.6.0",
+    "sphinx-sitemap>=2.6.0",
     "sphinx-toolbox~=4.0",
     "sphinx-lint==1.0.0",
     "sphinxcontrib-details-directive",
--- a/common.mk
+++ b/common.mk
@@ -221,11 +221,11 @@ test-find-slow:  ##- Identify slow tests. Set cutoff time in seconds with SLOW_C
 
 .PHONY: docs
 docs:  ## Build documentation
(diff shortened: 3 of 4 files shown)


docs/conf.py
```diff
@@ -53,10 +53,20 @@
...
 }
 
+# Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
+html_baseurl = "https://canonical-imagecraft.readthedocs-hosted.com/"
+
+if "READTHEDOCS_VERSION" in os.environ:
+    version = os.environ["READTHEDOCS_VERSION"]
+    sitemap_url_scheme = "{version}{link}"
+else:
+    sitemap_url_scheme = "latest/{link}"
```
medubelko, line 63: Outside of RTD, `latest` has no meaning, so we shouldn't be hard-coding it.

Two options here. One is that we remove the version in the sitemap for now, since Imagecraft's versions are ugly:

```suggestion
    sitemap_url_scheme = "{link}"
```

> https\://canonical-imagecraft.readthedocs-hosted.com/common/craft-parts/explanation/filesets.html

Or, we bring in the app's version, like on the other projects, warts and all, and stick it in the paths:

```diff
diff --git a/docs/conf.py b/docs/conf.py
index 3c2fd85..ebba976 100644
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,6 +21,7 @@ import sys
 
 import craft_parts_docs  # type: ignore
 
+import imagecraft
 
 project_dir = pathlib.Path(__file__).parent.parent.resolve()
 sys.path.insert(0, str(project_dir.absolute()))
@@ -35,7 +36,7 @@ sys.path.append(str(library_dir.absolute()))
 
 project = "Imagecraft"
 author = "Canonical"
-
+release = imagecraft.__version__
 copyright = "2023-%s, %s" % (datetime.date.today().year, author)
 
 # region Configuration for canonical-sphinx
@@ -58,9 +59,10 @@ html_baseurl = "https://canonical-imagecraft.readthedocs-hosted.com/"
 
 if "READTHEDOCS_VERSION" in os.environ:
     version = os.environ["READTHEDOCS_VERSION"]
-    sitemap_url_scheme = "{version}{link}"
 else:
-    sitemap_url_scheme = "latest/{link}"
+    version = release
+
+sitemap_url_scheme = "{version}{link}"
 
 extensions = [
     "canonical_sphinx",
```

> https\://canonical-imagecraft.readthedocs-hosted.com/**0.0.post692+ge989e59**/common/craft-parts/explanation/filesets.html

Given the state of Imagecraft's versioning, the former is certainly prettier. But in other apps, we might find it useful to plug the version into the sitemap when testing outside RTD.

I'm fine with either. What do you think?

docs/conf.py
```diff
@@ -53,10 +53,20 @@
     "source_edit_link": "https://github.com/canonical/imagecraft",
 }
 
+# Sitemap configuration: https://sphinx-sitemap.readthedocs.io/
+html_baseurl = "https://canonical-imagecraft.readthedocs-hosted.com/"
+
+if "READTHEDOCS_VERSION" in os.environ:
+    version = os.environ["READTHEDOCS_VERSION"]
+    sitemap_url_scheme = "{version}{link}"
```
medubelko, line 61: I don't know whether the current guidance for adding a meta-sitemap for versions is fully hashed-out, but let's see if this gets us where we want.

pyproject.toml
```diff
@@ -47,6 +47,7 @@ docs = [
     "pydantic-kitbash==0.0.5",
     "sphinx-autobuild~=2024.2",
     "sphinx-pydantic==0.1.1",
+    "sphinx-sitemap==2.6.0",
```
medubelko, line 50: Not rolling?

medubelko (changes requested):
Looks like it's working!

https://canonical-imagecraft--158.com.readthedocs.build/158/sitemap.xml

Let's look at the slug for local builds, then I should think this should be ready.

pyproject.toml
jahn-junior: Oops, that wasn't intentional.

https://github.com/canonical/imagecraft/pull/158/commits/2d015e1e7c4588da4acd0e60586ac5e9697c17b9

docs/conf.py
```diff
@@ -163,7 +173,7 @@
 
 
 def generate_cli_docs(nil):
-    gen_cli_docs_path = (project_dir / "tools" / "docs" / "gen_cli_docs.py").resolve()
+    gen_cli_docs_path = (project_dir / "tools/docs/gen_cli_docs.py").resolve()
```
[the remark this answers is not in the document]
jahn-junior: Nope, it just bothered me that all of the other paths in the file used a single string.

medubelko:
> Would it be better to just drop `latest` from local builds? It seems unnecessary now that you point it out.

It is, except we currently keep the `latest` and `stable` bits in our schemes. So the URLs would all be 404 if anyone ever checked them locally.

We can do away with the branch name for the default builds (in other words, make https://canonical-imagecraft.readthedocs-hosted.com/how-to-guides/basic_image/ reachable), but we'll need to make the switch first. But for technical reasons I don't think we're ready to make that change.

jahn-junior:
@medubelko If we want to keep the URLs valid in the local sitemaps, should I also update `make docs` to use the `dirhtml` builder? Otherwise, the local sitemap URLs will all have the `.html` extension.

medubelko (approved):
@jahn-junior Good find on the bug and the workaround!

PR #183, opened 2025-07-18
feat: add configure hook

### LINKED ISSUE #182: Add a `configure` snap hook
### What needs to get done

Add a `configure` snap hook to enable switching between multipass and lxd as the default provider.

Also update the docs to replace the use of the env var by a call to `snap set imagecraft provider=multipass`.

### Why it needs to get done

Imagecraft cannot yet fully run in lxd (cannot execute the pack step in it). So configuring the snap would improve user experience.

upils:

---

Add a configure snap hook to configure the default provider by calling `snap set imagecraft provider=multipass`.

Heavily inspired from https://github.com/canonical/snapcraft/pull/3898

IMAGECRAFT-35
Fixes #182

### DIFF
--- a/docs/how-to-guides/basic_image.rst
+++ b/docs/how-to-guides/basic_image.rst
@@ -94,7 +94,8 @@ The packing can be run in two different environments:
 
   .. code-block:: bash
 
-      CRAFT_BUILD_ENVIRONMENT=multipass imagecraft --verbosity debug pack
+      sudo snap set imagecraft provider=multipass
+      imagecraft --verbosity debug pack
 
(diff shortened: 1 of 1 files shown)


upils:
> What's the advantage of doing it this way over using the `ConfigService`? Make an [`ImagecraftConfigModel`](https://github.com/canonical/craft-application/blob/main/craft_application/_config.py#L24-L38) and add `provider` to it, and then it can be configured using snap configuration and environment variables.

Thanks for pointing that out, I completely missed this service. It looks like this  even simpler than what you suggested thanks to the existing  `SnapConfig` already handling the `provider`.

PR #191, opened 2025-08-05
docs: add filesystems model to reference

upils:

---

IMAGECRAFT-41

### DIFF
--- a/docs/reference/imagecraft-yaml-file.rst
+++ b/docs/reference/imagecraft-yaml-file.rst
@@ -50,6 +50,8 @@ alongside operational values such as its supported architectures and build envir
 .. kitbash-field:: Project volumes
     :override-type: dict[str, Volume]
 
+.. kitbash-field:: Project filesystems
+    :override-type: dict[str, FilesystemMount]
 
 Part keys
 ---------
@@ -232,3 +234,37 @@ The following keys can be declared for each partition listed in the ``structure`
 
 .. kitbash-field:: StructureItem filesystem_label
     :prepend-name: volumes.<volume-name>.structure.<partition>
+
+
+Filesystem keys
+---------------
+
+The following keys can be declared for each filesystem mount listed.
+
+.. py:currentmodule:: craft_parts.filesystem_mounts
+
+.. kitbash-field:: FilesystemMountItem mount
+    :prepend-name: filesystems.<filesystem-name>.<filesystem>
+
+**Examples**
+
+.. code-block:: yaml
+
+    mount: "/"
+
+.. code-block:: yaml
+
+    mount: "/boot/efi"
+
+.. kitbash-field:: FilesystemMountItem device
+    :prepend-name: filesystems.<filesystem-name>.<filesystem>
+
+**Examples**
+
+.. code-block:: yaml
+
+    device: "(default)"
+
+.. code-block:: yaml
+
+    device: "(volume/pc/rootfs)"


jahn-junior (commented):
Thanks for keeping this up-to-date! Could we work together to come up with some descriptions for these while we're here? We can just add them to the directive's content as of Kitbash 0.0.7:

```
.. kitbash-field:: <model-name> <key-name>

    <description>
```

upils:
> Thanks for keeping this up-to-date! Could we work together to come up with some descriptions for these while we're here? We can just add them to the directive's content as of Kitbash 0.0.7:
> 
> ```
> .. kitbash-field:: <model-name> <key-name>
> 
>     <description>
> ```

@jahn-junior Yes! I proposed some short description. Let me know if you think it should be more detailed.

However when trying your suggestion I still get `None` as the first line of the description before my content.

image

If I remove the indentation, the `Description` title is gone. Am I doing something wrong or is that a bug?

docs/reference/imagecraft-yaml-file.rst
```diff
@@ -232,3 +234,37 @@ The following keys can be declared for each partition listed in the ``structure`
...
+
+Filesystem keys
+---------------
+
+The following keys can be declared for each filesystem mount listed.
+
+.. py:currentmodule:: craft_parts.filesystem_mounts
+
+.. kitbash-field:: FilesystemMountItem mount
+    :prepend-name: filesystems.<filesystem-name>.<filesystem>
```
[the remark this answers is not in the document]
jahn-junior: `<filesystem-mount>` feels redundant given that its parent is a named filesystem. I lean towards `<mount>` because the hierarchy already asserts that it belongs to a filesystem. It's also consistent with how the `volumes` child keys are titled.

docs/reference/imagecraft-yaml-file.rst
```diff
@@ -232,3 +234,41 @@ The following keys can be declared for each partition listed in the ``structure`
...
+    mount: "/"
+
+.. code-block:: yaml
+
+    mount: "/boot/efi"
+
+.. kitbash-field:: FilesystemMountItem device
+    :prepend-name: filesystems.<filesystem-name>.<filesystem>
+
+    Device to be mounted. Must reference a partition.
```
jahn-junior, line 264: Sorry about this. I had only planned on this being used to supplement the description, so I didn't account for the case where there is no field description. I created an [issue](https://github.com/canonical/pydantic-kitbash/issues/56) to track this.

In the meantime, we can just add these in with rST.

```suggestion
**Description**

Device to be mounted. Must reference a partition.
```

We do the same thing in Snapcraft for application-specific behavior.

jahn-junior: Suggestion for the description itself:

```suggestion
The device to be mounted. This must reference one of the partitions defined in `volumes.<volume-name>.structure`.
```

docs/reference/imagecraft-yaml-file.rst
```diff
@@ -232,3 +234,41 @@ The following keys can be declared for each partition listed in the ``structure`
...
+---------------
+
+The following keys can be declared for each filesystem mount listed.
+
+.. py:currentmodule:: craft_parts.filesystem_mounts
+
+.. kitbash-field:: FilesystemMountItem mount
+    :prepend-name: filesystems.<filesystem-name>.<filesystem>
+
+    Mountpoint to mount the device at.
```
jahn-junior, line 249:
```suggestion
    The device's mount point.
```

PR #322, opened 2026-04-20
feat: support 26.04 as a build-base

bepri:
Adds 26.04 to the list of supported `build-base` values. I didn't change any defaults here or update the main tests here as I'm not sure we want it to be the default yet.

---


imagecraft/models/project.py
```diff
@@ -67,7 +67,14 @@ def _validate_platform_set(self) -> Self:
 
 BaseT = Literal["bare"]
 BuildBaseT = typing.Annotated[
-    Literal["ubuntu@20.04", "ubuntu@22.04", "ubuntu@24.04", "ubuntu@25.10", "devel"],
+    Literal[
+        "ubuntu@20.04",
+        "ubuntu@22.04",
+        "ubuntu@24.04",
+        "ubuntu@25.10",
```
bepri, line 74: Why is this here? It isn't documented

spread.yaml
```diff
@@ -152,12 +159,7 @@ prepare: |
   snap install yq --devmode  # Disable confinement so we can read files from anywhere
 
   # Setup lxd
-  # older linux releases have separate packages for lxd and lxc (lxd-client)
```
bepri, line 155: Fly-by because we weren't actually testing on 20.04 anymore

imagecraft/models/project.py
lengau: Because we were testing on 25.10 before 26.04 was released so it's still in the code.

lengau (approved):
Thanks!

smethnani (approved):

PR #332, opened 2026-04-28
fix(snap-prepare): make keys optional

smethnani:
This PR:
- makes only one of `snap-preseed-snaps` or `snap-preseed-model-assert` required
- changes default validation to "enforce"

Currently `snap-preseed-snaps` is required for the snap-preseed plugin but the user can provide a model assertion file which would contain a list of snaps.  

`snap prepare-image` sets the validation mode to `enforce` by default.

---

### DIFF
--- a/imagecraft/plugins/snap_preseed_plugin.py
+++ b/imagecraft/plugins/snap_preseed_plugin.py
@@ -20,8 +20,8 @@ import shlex
 from typing import Literal, cast
 
 from craft_parts.plugins import Plugin, PluginProperties
-from pydantic import field_validator
-from typing_extensions import override
+from pydantic import field_validator, model_validator
+from typing_extensions import Self, override
 
 from ._utils import resolve_snap, validate_snap_refs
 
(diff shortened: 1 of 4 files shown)


imagecraft/plugins/snap_preseed_plugin.py
```diff
@@ -31,14 +31,23 @@ class SnapPreseedPluginProperties(PluginProperties, frozen=True):
...
     snap_preseed_channel: str | None = None
     snap_preseed_model_assert: str = ""
-    snap_preseed_validation: Literal["ignore", "enforce"] = "ignore"
+    snap_preseed_validation: Literal["ignore", "enforce"] | None = None
     snap_preseed_assertions: list[str] = []
     snap_preseed_revisions: str | None = None
     snap_preseed_write_revisions: str | bool = False
 
+    @model_validator(mode="after")
+    def _snaps_or_model_assert(self) -> Self:
```
lengau, line 43: Sad that we have to do it this way, but I don't see a pydantic-native way to make mutually exclusive fields that will also get sent to the JSON schema.

If it's super easy could you [add to the JSON schema of this model](https://stackoverflow.com/a/28172831) that these are mutually exclusive? If it takes more than ~5 minutes though let's instead just make it a task for later.

lengau (approved):
LGTM, thanks!

mr-cal (approved):

imagecraft/plugins/snap_preseed_plugin.py
smethnani: They're not mutually exclusive. You can have a model assertion file and also add additional `--snap=snap-name` args. 

Maybe "**At least** one of snap-preseed-snaps or snap-preseed-model-assert must be set" is a better message

mr-cal: I like that suggested wording better.

PR #333, opened 2026-04-28
docs: add snap-preseed plugin

smethnani:
Adds a reference doc for the snap-preseed plugin

---

- [x] I've added or updated any relevant documentation.


docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,102 @@
+.. _reference-snap-preseed-plugin:
+
+Snap Preseed Plugin
+===================
+
+The snap-preseed plugin seeds snaps into a Classic Ubuntu image by running ``snap
```
lengau:
```suggestion
The snap-preseed plugin seeds snaps into a Classic image by running ``snap
```

Gonna defer to @jahn-junior here, but I'd prefer if possible to not limit this to Ubuntu. We have known reasons for Imagecraft to need to support at least one non-Ubuntu distro, and this would apply to any non-core distro with snapd installed.

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,102 @@
...
+
+snap-preseed-snaps
+~~~~~~~~~~~~~~~~~~
+
+**Type** list of strings
+
+**Required**
+
+The snaps to seed into the image. Each entry can be a snap name or a path to a local
+snap.
```
lengau: IIRC these values can be:

- A store name
- A store name + channel (e.g. `lxd/latest/edge`)
- A local path *within the project directory*

(Technically the emphasised part of the last isn't strictly true but...)

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,102 @@
...
+
+
+snap-preseed-channel
+~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The default channel to use when fetching snaps from the store.
```
lengau: A channel in a snap above overrides this, right?

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,102 @@
...
+
+
+snap-preseed-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The path to a model assertion file that defines the Ubuntu Classic image.
```
lengau: I'd like to workshop this a bit as well. Technically the assertion file for classic mode only defines limited stuff about the snaps, right?

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,102 @@
...
+prepare-image --classic``, making them available on first boot.
+
+
+Keys
+----
+
+This plugin provides the following unique keys.
+
+
+snap-preseed-snaps
```
lengau, line 33: Should these be kitbash directives pointing to the plugin properties model?

docs/reference/plugins/snap_preseed_plugin.rst
smethnani: Yes that's correct

docs/reference/plugins/snap_preseed_plugin.rst
smethnani: ~~True, hows: `The path to a model assertion file that defines the snaps to seed in the image.`~~

Actually nevermind. For classic systems, the model assertion doesn't need to include snaps at all. It could just have brand and model info. The only other difference I see is `classic` should be set to `true` and there shouldn't be a kernel snap present.

docs/reference/plugins/snap_preseed_plugin.rst
smethnani: Went ahead and removed it. Let me know what you think @jahn-junior!

jahn-junior: I would've suggested the same thing. Thanks to both of you!

docs/reference/plugins/snap_preseed_plugin.rst
jahn-junior: Eventually, yes. We have an epic for this next cycle, so let's table it for now and tackle it as part of a larger initiative.

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,112 @@
+.. _reference-snap-preseed-plugin:
+
+Snap Preseed Plugin
+===================
```
jahn-junior:
```suggestion
Snap Preseed plugin
===================
```

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,112 @@
...
+
+The snap-preseed plugin seeds snaps into a Classic image by running ``snap prepare-image
```
jahn-junior:
```suggestion
The Snap Preseed plugin seeds snaps into a Classic image by running ``snap prepare-image
```

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,112 @@
...
+
+snap-preseed-snaps
+~~~~~~~~~~~~~~~~~~
+
+**Type** list of strings
+
+The snaps to seed into the image. Valid entries are:
+
+- a snap name
+- a snap name and channel in the format ``snap-name/channel``
```
jahn-junior: Nit
```suggestion
- a snap name and channel in the format ``<snap-name>/<channel>``
```

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,112 @@
...
+- a path to a local snap within the project directory
+
+
+snap-preseed-channel
+~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The default channel to use when fetching snaps from the store. This is overridden by any
+channel specified directly in a snap reference.
```
jahn-junior:
```suggestion
The default channel to use when fetching snaps from the store. This is overridden by snaps
listed with a channel in ``snap-preseed-snaps``.
```

jahn-junior (commented):
I'm going to hold off on reviewing this in more detail until we propagate any changes to the [UC Prepare plugin docs](https://github.com/canonical/imagecraft/pull/334/changes).

I already know I'll confuse myself if I try to do them in parallel.

docs/reference/plugins/snap_preseed_plugin.rst
smethnani: Forgot to add the channel precedence in the uc-prepare reference so added it there too.

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,132 @@
...
+
+
+snap-preseed-channel
+~~~~~~~~~~~~~~~~~~~~
+
+**Type:** string
+
+The default store channel to fetch snaps from, overriding any channels in the model
+assertion, if one is provided. For model assertion with a grade, the ``grade`` must be
+set to ``dangerous``.
```
smethnani: A model assertion for a classic image could have no grade (if it uses the older format without a snaps section), in which case no restrictions apply. This isn't the case for core images that use core20+ since they all must use the newer "extended" format which requires a grade (or defaults to "signed").

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,132 @@
+.. meta::
+    :description: Reference for the Snap Preseed plugin, which creates the seed directory
+                  for classic images.
+
+.. _reference-snap-preseed-plugin:
+
+Snap Preseed plugin
+===================
```
jahn-junior:
```suggestion
Snap-preseed plugin
===================
```
In the docs sync today, we decided to hyphenate two-word plugin names (e.g., "Cargo Use" to "Cargo-use").

We'll also need to change instances of "UC Prepare" to "uc-prepare". We can do that here or, if you'd like to preserve this PR's scope, I can open a separate one.

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,132 @@
...
+This is overridden by snaps listed with a channel in ``snap-preseed-snaps``.
+
+
+snap-preseed-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type:** string
+
+The path to the model assertion file that defines the target device. If this key is set,
+the model assertion should have ``classic`` set to ``true``.
```
jahn-junior:
```suggestion
The path to the model assertion file that defines the target device. If this key is set,
the model assertion's ``classic`` key must be set to ``true``.
```
This is subjective. I don't love the added inflection, but it feels slightly more natural to me.

docs/reference/plugins/snap_preseed_plugin.rst
jahn-junior: This detail would've gone over my head. Thanks for highlighting it.

One small suggestion for concision. This is a slightly ambiguous "it", but I think it's clear enough.

> If a model assertion is provided and has a grade, it must be set to ``dangerous``.

docs/reference/plugins/snap_preseed_plugin.rst
```diff
@@ -0,0 +1,132 @@
...
+
+**Type:** list of strings
+
+The snaps to seed into the image. Valid entries are:
+
+- a snap name
+- a snap name and channel in the format ``<snap-name>/<channel>``
+- a path to a local snap within the project directory
+
+If a model assertion with a grade of ``signed`` or ``secured`` is provided, only snaps
```
jahn-junior:
```suggestion
If the model assertion has a grade of ``signed`` or ``secured``, only snaps
```
Sorry, I just noticed an error with my suggestion for the [UC Prepare counterpart](https://documentation.ubuntu.com/imagecraft/latest/reference/plugins/uc_prepare_plugin/#uc-prepare-snaps). It goes from plural (model assertion**s**) to singular (**the** model assertion) in the same sentence. Similar to the name formatting, we can propagate this change here or I can open a separate PR to normalize this piece.

jahn-junior (changes requested):
Thanks for propagating the changes from the other doc!

I left a few minor suggestions. One is the result of a discussion earlier today and another is completely my fault. The rest are nits.

My main question is this: can we standardize the ordering of the keys common to this and the UC-prepare plugin? It feels weird to reference the model assertion before its associated key has been introduced on the page. Beyond that, I won't be too picky about the ordering. We can always revisit it later.

docs/reference/plugins/snap_preseed_plugin.rst
smethnani: Happy to make the uc-prepare changes here too :)

Lmk if they weren't propogated correctly

jahn-junior: Looks good. Thanks for doing that!

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -1,13 +1,13 @@
 .. meta::
-    :description: Reference for the UC Prepare plugin, which creates the seed directory
+    :description: Reference for the uc-prepare plugin, which creates the seed directory
                   for Ubuntu Core images.
 
 .. _reference-uc-prepare-plugin:
 
-UC Prepare plugin
+uc-prepare plugin
```
jahn-junior: I think we should stick with the old capitalization when referring to the plugin by name. It's more consistent with the product branding (**U**buntu **C**ore). Open to other opinions on this, though.


```suggestion
UC-prepare plugin
```

jahn-junior (approved):
Thanks so much! One last comment on the name, but it doesn't warrant a re-review.

lengau (approved):

PR #334, opened 2026-04-28
docs: add uc-prepare plugin

smethnani:
Adds a reference doc for the uc-prepare plugin

---

- [x] I've added or updated any relevant documentation.


docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+
+
+uc-prepare-preseed
+~~~~~~~~~~~~~~~~~~
+
+**Type** boolean
+
+**Default** ``false``
+
+If ``true``, the plugin runs the snap preseeding process to reduce first-time boot.
```
steinbro:
```suggestion
If ``true``, the plugin runs the snap preseeding process to reduce first-boot time.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Path to a sysfs overlay directory for preseeding. Requires ``uc-prepare-preseed`` to be
+enabled.
+
+
+Output
+------
+
+The seed content for the image is places under ``system-seed``. Use ``organize`` to
+place it in the appropriate partition.
```
steinbro:
```suggestion
The seed content for the image is placed under ``system-seed``. Use ``organize`` to
place it in the desired partition.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+
+
+uc-prepare-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+**Required**
+
+The path to the model assertion file that defines the Ubuntu core image.
```
steinbro: How would a user typically get one of these?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+
+
+uc-prepare-validation
+~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+**Default** ``ignore``
+
+The validation mode for snap signatures. Valid values are ``ignore`` and ``enforce``.
```
steinbro: How will the plugin behave in each of these modes?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+
+Additional assertion files to include in the image.
+
+
+uc-prepare-revisions
+~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+Path to a manifest file specifying snap revisions to use.
```
steinbro, line 93: What would such a file look like? Where would the user get one?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,135 @@
...
+The key to use for signing the preseed assertion. Requires ``uc-prepare-preseed`` to be
+enabled.
+
+
+uc-prepare-apparmor-features-dir
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+Path to the AppArmor features directory to use during preseeding.
```
steinbro: Is "features directory" a standard AppArmor term? Is there AppArmor documentation we can link to that describes what this would look like?

steinbro (commented):
I feel like this doc would only make sense to someone already very well-versed in Ubuntu Core images, perhaps even the specific arguments to `snap prepare-image`. Explaining these concepts in depth is beyond the scope of a reference like this, but I do think the user deserves a description of not just what the field is but how/where to find an appropriate value. This is especially true for fields that expect paths to files that come from other tools.

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: They would either create it or be provided with one (eg some official one)

Should we link [this page](https://documentation.ubuntu.com/core/tutorials/build-your-first-image/create-a-model/)?

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: There's [this AppArmor wiki](https://gitlab.com/apparmor/apparmor/-/wikis/AppArmorInterfaces/#syskernelsecurityapparmorfeatures)

For `snap prepare-image`, I only know of the [manpages](https://manpages.debian.org/testing/snapd/snap.8.en.html#prepare-image). 

There is a bit more detail for AppArmor features in the [preseeding section of this page](https://documentation.ubuntu.com/core/how-to-guides/image-creation/optimise-boot-speed/#preseeding). Perhaps worth linking on this reference page too, or just mentioning:

>A custom AppArmor features directory may be specified with --apparmor-features-dir=.... The target should be a snapshot of sys/kernel/security/apparmor/features from the target system. If not specified, the sys/kernel/security/apparmor/features from the host system will be used.

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: It's a text file with a list of snap names and revision numbers like this:
```
core24 1587
snapd 26865
```
The user could either write it themselves or use the output from a previous `snap prepare-image --write-revisions` command, which would write the downloaded snap revisions to a file that can then be used as the revisions input

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: For plugin behaviour, the difference is the command run: `snap prepare-image --validation={ignore/enforce}`.

From snapd's side, it controls validation-set enforcement

docs/reference/plugins/uc_prepare_plugin.rst
steinbro: I would say so, but I defer to our TAs as to what's proper practice here.

docs/reference/plugins/uc_prepare_plugin.rst
steinbro: Right, so I feel like we should say "When set to `enforce`, the build will fail if the signatures don't match" if that is indeed true.

docs/reference/plugins/uc_prepare_plugin.rst
steinbro: I think even that trivial example is worth specifying.

docs/reference/plugins/uc_prepare_plugin.rst
steinbro: I think that last block comment is clear and unambiguous. Perhaps make "AppArmor features" link to the wiki.

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: I believe it's controlling enforcement for validation set constraints rather than signatures. I've fixed the text to reflect that

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-revisions
+~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+Path to a manifest file specifying snap revisions to use.
+
+The following is an example manifest file:
```
steinbro:
```suggestion
Each line in a manifest file identifies a snap by name and revision number, like so:
```

docs/reference/plugins/uc_prepare_plugin.rst
steinbro: The link is helpful, but the meaning of the parameter is still ambiguous. Are we configuring a property of the build or of the resulting image? That is, when would the validation that is enforced or ignored actually performed?

smethnani: It happens during the `snap prepare-image` call, I believe when the snaps are fetched from the store

steinbro: But the doc you linked to also says "Enforced validation sets prevent listed snaps and components from being removed from the system." This suggests that the runtime behavior of the resulting image will also be different, no?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+uc-prepare-validation
+~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+**Default** ``enforce``
+
+Controls whether `validation set
+<https://snapcraft.io/docs/explanation/how-snaps-work/validation-sets>`_ constraints are
+enforced. Valid values are ``ignore`` and ``enforce``.
```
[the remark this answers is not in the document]
smethnani: The default should be "enforce"

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
+.. _reference-uc-prepare-plugin:
+
+UC Prepare Plugin
+=================
```
jahn-junior:
```suggestion
UC Prepare plugin
=================
```
We use sentence case for titles.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+The uc-prepare plugin prepares snaps for Ubuntu Core images using the ``snap
```
jahn-junior:
```suggestion
The UC Prepare plugin prepares snaps for Ubuntu Core images with the ``snap
```
>  uc-prepare -> UC Prepare

We establish the more natural name in the title but then immediately revert. Let's keep it consistent.

> using the ``snap prepare-image`` command -> with the ``snap prepare-image`` command

This one is more nitpicky, but I like to avoid introducing extra verbs. I especially like to avoid "use" because it's so nondescript. There are exceptions to this, but it's a good thing to keep in mind.

Also, it's slightly outside our purview, but I think we should _briefly_ describe what the ``prepare-image`` command does. The verb "preseed" seems like jargon to me.

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: To keep a more consistent tone, I think it'd be better to link to [the model assertion reference](https://documentation.ubuntu.com/core/reference/assertions/model/).

Feel free to play around with the specifics, but I'm thinking something like:

> For a complete reference of the contents of a model assertion, refer to [ Ubuntu Core | model]().

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+**Required**
+
+The path to the model assertion file that defines the Ubuntu Core image.
```
jahn-junior: Doesn't the model assertion define the target device?

As a new user reading this, I'd be thinking, "I thought the imagecraft.yaml file defined the image".

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+<https://documentation.ubuntu.com/core/tutorials/build-your-first-image/create-a-model>`_
+tutorial from the Ubuntu Core documentation contains more details.
+
+
+uc-prepare-snaps
+~~~~~~~~~~~~~~~~
+
+**Type** list of strings
+
+Additional snaps to seed into the image. Each entry can be a snap name or a path to a
```
jahn-junior: > Additional

In addition to what? This implies that there are snaps seeded into the image by default. If so, where do they come from?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+local snap.
+
+
+uc-prepare-channel
+~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The default channel to use when fetching snaps from the store.
```
jahn-junior: Since this isn't required, I assume it has a default?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+<https://snapcraft.io/docs/explanation/how-snaps-work/validation-sets>`_ constraints are
+enforced. Valid values are ``ignore`` and ``enforce``.
+
+
+uc-prepare-assertions
+~~~~~~~~~~~~~~~~~~~~~
+
+**Type** list of strings
+
+Additional assertion files to include in the image.
```
jahn-junior, line 82: I think we should link to the assertions index here.

> For a complete list of available assertions, refer to ...

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-revisions
+~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+Path to a manifest file specifying snap revisions to use.
```
jahn-junior, line 93: I assume these should match the snaps listed with `uc-prepare-snaps`. Is this a hard constraint? What happens if I don't list one of those snaps or list one that isn't present?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-write-revisions
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string or boolean
+
+If ``true``, the plugin writes a manifest file with the resolved snap revisions to
+``seed.manifest``. If a string, it is treated as the path to which the manifest will be
+written.
```
jahn-junior: What's the default?

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-preseed
+~~~~~~~~~~~~~~~~~~
+
+**Type** boolean
+
+**Default** ``false``
+
+If ``true``, the plugin runs the snap preseeding process to reduce first-boot time.
```
jahn-junior: > snap preseeding process

We need to establish what "preseeding" is earlier on the page or use more descriptive language here.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+
+uc-prepare-preseed-sign-key
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The key to use for signing the preseed assertion. Requires ``uc-prepare-preseed`` to be
```
jahn-junior:
```suggestion
The signing key for the preseed assertion. Requires ``uc-prepare-preseed`` to be
```
To avoid overloading 'key'

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+**Type** string
+
+Path to the `AppArmor features
+<https://gitlab.com/apparmor/apparmor/-/wikis/AppArmorInterfaces/#syskernelsecurityapparmorfeatures>`_
+directory to use during preseeding.
+
+This directory should be a snapshot of ``sys/kernel/security/apparmor/features`` from
+the target system. If not specified, the ``sys/kernel/security/apparmor/features`` from
+the host system will be used when ``uc-prepare-preseed`` is enabled.
```
jahn-junior: General language tweaks
```suggestion
This directory should be a snapshot of the ``sys/kernel/security/apparmor/features``
file from the target system. If this key is unset and ``uc-prepare-preseed`` is set to
``true``, the file is pulled from the host system.
```
> ``uc-prepare-preseed`` is enabled

The _behavior_ can be enabled, but the key itself isn't toggled in any way. There are a few other instances of this in the other descriptions.


> when ``uc-prepare-preseed`` is enabled

Feels weird to refer to a sequence for a declarative file. It either is or isn't enabled.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,154 @@
...
+
+Path to a sysfs overlay directory for preseeding. Requires ``uc-prepare-preseed`` to be
+enabled.
+
+
+Output
+------
+
+The seed content for the image is placed under ``system-seed``. Use ``organize`` to
+place it in the desired partition.
```
jahn-junior:
```suggestion
The seed content for the image is placed under ``system-seed``. Place it in
the desired partition with the :ref:`PartSpec.organize_files` key.
```

jahn-junior (changes requested):
Thanks! I suggested a few language tweaks and raised questions around a few potential gaps.

I'll do another pass when our services are back and I can view the rendered page.

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Added a little more description in the intro and introduced the "preseed" jargon, what do you think?

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: The snaps come from the model assertion file. 

Wasn't sure if this was too much detail but I added that you can:
- add arbitrary snaps and local snaps only if the model assertion grade is dangerous
- add snaps already listed in the model assertion if the grade is higher (for example if they are listed as optional)

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Thanks! I tweaked this slightly as features is a directory and not a file

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+=================
+
+The UC Prepare plugin creates the seed directory for Ubuntu Core images with the ``snap
+prepare-image`` command. This downloads snaps and their associated assertions from the
+Snap store, organizing them into a structure that ``snapd`` reads at first boot to
+install to the system.
+
+The plugin can also preseed the image, which performs administrative tasks for the
+snaps being seeded during the image-building phase rather than during the initial boot,
+reducing first-boot time.
```
jahn-junior:
```suggestion
The plugin can also install snaps and run their hooks when the image is being packed.
This process, known as *preseeding*, reduces the initial boot time.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+
+
+uc-prepare-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+**Required**
+
+The path to the model assertion file that defines the target device, including its
```
jahn-junior:
```suggestion
The path to the model assertion file that defines the target device and its
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+required snaps.
+
+For a complete reference of the contents of a model assertion, refer to `Ubuntu Core |
+model <https://documentation.ubuntu.com/core/reference/assertions/model/>`_.
```
jahn-junior:
```suggestion
The Ubuntu Core documentation details model assertions and their fields in the `model
<https://documentation.ubuntu.com/core/reference/assertions/model/>`__ reference.
```
Revising my previous suggestion now that I have fresh eyes.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+entries are:
+
+- a snap name
+- a snap name and channel in the format ``<snap-name>/<channel>``
+- a path to a local snap within the project directory
+
+For model assertions with grades higher than ``dangerous`` (``signed``, ``secured``), only
+snaps already declared in the model assertion can be specified and must not be referenced
+by a local path. This is useful for including snaps that are listed as optional in the
+model assertion.
```
jahn-junior: I like this addition!
```suggestion
For model assertions with a grade of ``signed`` or ``secured``, only
snaps declared in the model assertion can be specified, and they can't be stored
locally. This is commonly used to include optional snaps from the model
assertion.
```
Let me know if you think we should change any of this wording.

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior:
```suggestion

The Ubuntu Core documentation lists the available assertion types in
`Assertions <https://documentation.ubuntu.com/core/reference/assertions/>`__.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+Keys
+----
+
+This plugin provides the following unique keys.
+
+
+uc-prepare-model-assert
+~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string
```
jahn-junior:
```suggestion
**Type:** string
```
These should all have colons. It's not super important since we'll refactor these with Kitbash next cycle, but it doesn't hurt to keep things consistent in the meantime.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+model assertion.
+
+
+uc-prepare-channel
+~~~~~~~~~~~~~~~~~~
+
+**Type** string
+
+The default channel to use when fetching snaps from the store. Requires the model
+assertion grade to be set to ``grade: dangerous``.
```
jahn-junior:
```suggestion
The default store channel to fetch snaps from. If this key is set,
the model assertion's ``grade`` must be set to ``dangerous``.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+
+.. _reference-uc-prepare-plugin:
+
+UC Prepare plugin
+=================
+
+The UC Prepare plugin creates the seed directory for Ubuntu Core images with the ``snap
+prepare-image`` command. This downloads snaps and their associated assertions from the
+Snap store, organizing them into a structure that ``snapd`` reads at first boot to
+install to the system.
```
jahn-junior:
```suggestion
prepare-image`` command. This downloads snaps and their assertions from the
Snap Store and prepares them to be installed during the initial boot.
```
I think this is a little more punchy, which is important for the first paragraph.

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: I think that context will be very appreciated by users outside Canonical. I left a follow-up suggestion on the specific wording.

https://github.com/canonical/imagecraft/pull/334/changes#r3192219722

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: And then a separate question for my understanding:

Why is it that in `uc-prepare-snaps`, additional snaps can't be local in `dangerous` model assertions, but this key requires a `dangerous` grade to specify the channel for remote snaps?

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: Thanks for adding the required grade. Would it be prudent add something like:

> For model assertions with more stable grades, the channel defaults to ...

I'm not up to speed on the specific behavior.

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+uc-prepare-write-revisions
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**Type** string or boolean
+
+**Default:** ``False``
+
+If ``true``, the plugin writes a manifest file with the resolved snap revisions to
+``seed.manifest``. If a string, it is treated as the path to which the manifest will be
+written.
```
jahn-junior:
```suggestion
If set to ``true``, the plugin writes the resolved snap revisions to the
``seed.manifest`` file. If set to a file path, the revisions are written there instead.
```

docs/reference/plugins/uc_prepare_plugin.rst
```diff
@@ -0,0 +1,176 @@
...
+
+
+uc-prepare-preseed
+~~~~~~~~~~~~~~~~~~
+
+**Type** boolean
+
+**Default** ``false``
+
+If ``true``, the plugin runs the snap preseeding process to reduce first-boot time.
```
jahn-junior:
```suggestion
If set to ``true``, the plugin preseeds snaps to improve the initial boot speed.
```

jahn-junior (changes requested):
The rendered page looks good! Just a handful of additional language tweaks

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: `dangerous` model assertions *can* use local snaps, `signed` and `secured` can't.

The grade constraints are from snapd (the `snap prepare-image` command). 

As far as I understand, model assertions are the source of truth for what is to be installed on the device, they are signed and shouldn't be changed. So adding additional snaps not listed in there is not permitted by snapd. Local snaps means they are not in the store and can't have their provenance checked or validated

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Only concern is "can't be stored locally" sounds like the snaps can't exist locally. The constraint is just that you must reference them from the store and not by a local path. 

I changed it to "they can't be referenced by a local path" but do you prefer something like: "and they can't be local snap files"?

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Woops. Thanks!

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Added the link. Did you mean to remove the "`Additional assertion files to include in the image.`" line in your suggestion?

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: The plugin doesn't set a default for this value but the model assertion has a `default-channel` field for each snap which snapd sets to `latest/stable` if not set explicitly (this is also in the model reference we link above)

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: The file isn't required. If you don't list a specific revision then the latest revision from the channel specified in the model assertion is fetched (those default to `latest/stable`if not specified). If you list a snap that isn't present, it's ignored.

lengau (approved):
Thanks!

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: The "Enforced validation sets prevent listed snaps and components from being removed from the system." refers to having the "enforce" in the model assertion's validation-sets. That enforcement is controlled by the model assertion file and can't be changed by `snap prepare-image`.

As far as I understood, `--validation=ignore` is supposed to control whether snapd checks validation-sets after downloading the snaps but it seems to still validate regardless of the validation mode its using. I've asked for clarification from the snapd team.

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: Thanks! I think I get it now.

So, any time a snap's provenance doesn't match the model assertion (whether because it's sourced locally or from a different channel), the grade needs to be `dangerous`.

docs/reference/plugins/uc_prepare_plugin.rst
jahn-junior: > Did you mean to remove the "Additional assertion files to include in the image." line in your suggestion?

Nope! You interpreted the suggestion correctly :)

jahn-junior (approved):
LGTM. Thanks for your work on this!

docs/reference/plugins/uc_prepare_plugin.rst
smethnani: Will update this in a separate PR if this is not a bug from the snapd side.

PR #341, opened 2026-05-06
fix: clear ty diagnostics for partition-number helpers

mwhudson:
ty (the static type checker run by `make lint-ty`) flagged three genuine issues in code introduced by the MBR support PR:

- grubutil._part_num used `# type: ignore[union-attr]` (mypy syntax) to reach `partition_number` on a union containing variants that don't define it. Capture the result of getattr() in a local instead, removing the need for any ignore.
- ImageService._get_partition_numbers was annotated with `GPTVolume | MBRVolume`, but its only caller passes `GPTVolume | MBRVolume | HybridVolume`. The implementation already handles HybridVolume correctly (volume_schema is HYBRID, so the MBR-extended path is skipped); widen the parameter type to match.
- test_part_num_{mbr_plain,mbr_extended,gpt} build their structure list from MagicMock instances. ty doesn't see through `spec=MBRStructureItem`, so the call to _part_num looked ill-typed. Cast to MBRStructureList / GPTStructureList at construction so the rest of the test reads naturally.

`make lint-ty` and `make lint-ruff` both pass clean afterwards.

### DIFF
--- a/imagecraft/pack/grubutil.py
+++ b/imagecraft/pack/grubutil.py
@@ -231,8 +231,9 @@ def _part_num(name: str, structure: StructureList) -> int | None:
     )
     for i, structure_item in enumerate(structure):
         if structure_item.name == name:
-            if getattr(structure_item, "partition_number", None) is not None:
-                return structure_item.partition_number  # type: ignore[union-attr]
+            explicit = getattr(structure_item, "partition_number", None)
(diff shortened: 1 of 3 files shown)


smethnani (approved):
Thanks!

lengau (approved):
Thanks!

PR #348, opened 2026-05-24
No loops for grub

mwhudson:
Part 2 of my "no loop device" effort
---


lengau:
This looks interesting. I wonder if, with this approach, we'd be able to remove special handling for grub altogether and just make grub a parts plugin?

mwhudson:
> This looks interesting. I wonder if, with this approach, we'd be able to remove special handling for grub altogether and just make grub a parts plugin?

I don't really know what a plugin can do but I don't think it can dd the content into the image file directly? I guess an interface can be added to allow it to do so but is it worth it? I guess the way to answer that is to go look for what it takes to be bootable on other machines. I know some ARM boards require dd-ing the second stage bootloader to a fixed offset and things like that but it's hard to speculate without use cases.

I spent a while talking to a robot about this and it made the commit i've just pushed.

PR #375, opened 2026-07-10
test: add core boot spread test

smethnani:
- Adds spread test suite for booting core images (only includes arm64 image)
- Marks boot tests as manual
- Also updates qemu arguments for booting classic arm64 image to run faster

---


tests/spread/boot/core/sentinel/bin/sentinel.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+if [ "$(uname -m)" = "x86_64" ]; then
+  SERIAL_CONSOLE="/dev/ttyS0"
+else
+  SERIAL_CONSOLE="/dev/ttyAMA0"
+fi
+echo "HELLO FROM IMAGECRAFT" > $SERIAL_CONSOLE
```
steinbro:
```suggestion
echo "HELLO FROM SENTINEL SNAP" > $SERIAL_CONSOLE
```

tests/spread/boot/core/task.yaml
```diff
@@ -0,0 +1,55 @@
...
+    $QEMU_ARGS \
+    -accel kvm -accel tcg,thread=multi \
+    -smp "$(nproc)" \
+    -nographic \
+    -m 1G \
+    -drive if=pflash,format=raw,readonly=on,file=$UEFI_CODE \
+    -drive if=pflash,format=raw,file=/tmp/uefi_vars.fd \
+    -drive file=pc.img,format=raw,index=0,if=virtio,cache=unsafe 2>&1 | tee output.log || true
+
+  MATCH "HELLO FROM IMAGECRAFT" output.log
```
steinbro:
```suggestion
  MATCH "HELLO FROM SENTINEL SNAP" output.log
```

tests/spread/boot/core/sentinel/bin/sentinel.sh
```diff
@@ -0,0 +1,13 @@
...
+
+sleep 5
+dbus-send --system --print-reply \
+        --dest=org.freedesktop.login1 /org/freedesktop/login1 \
+        "org.freedesktop.login1.Manager.PowerOff" boolean:true
```
steinbro, line 13: I'm curious, why the D-Bus message instead of the `shutdown` command?

tests/spread/boot/core/arm64/imagecraft.yaml
```diff
@@ -0,0 +1,55 @@
...
+    source: .
+    uc-prepare-model-assert: model.assert
+    uc-prepare-snaps:
+      - sentinel.snap
+    organize:
+      "system-seed/*": (volume/pc/ubuntu-seed)
+    prime:
+      - -kernel
+      - -gadget
+      - -resolved-content
```
steinbro, line 30: Should the uc-prepare plugin automatically remove these items after building? Will there ever be a situation where these should be kept in the artifact?

steinbro (approved):

tests/spread/boot/core/sentinel/bin/sentinel.sh
smethnani: Shutdown command was being ignored. [This discussion](https://forum.snapcraft.io/t/shutdown-ubuntu-core-running-in-chroot-snap-classic/6441/6) made me try dbus 🤷

tests/spread/boot/core/arm64/imagecraft.yaml
smethnani: Yes it should. I don't know of a situation where we would need them and in fact the symlinks in `kernel` and `resolved-content` do harm. Will open another PR for that. The solution should be to use the build dir as the target for the prepare-image command and cp only the system-seed directory to the install dir

lengau: Agreed. *If* we find a case we can make "actually do stage these" an opt-in option.

lengau (approved):
This is great, thank you!

PR #388, opened 2026-07-31
docs: add custom files how-to

smethnani:
Adds how-to docs for including custom-files

---


docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+
+
+Copy the files to the image
+---------------------------
+
+Copy the files to the image with a new part that uses the :ref:`dump plugin
+<craft_parts_dump_plugin>` and the :ref:`organize <PartSpec.organize_files>` key. The
+part should run after the root file system is in place.
+
+For example, to copy the script ``my-files/my-script.sh`` to ``/usr/local/bin/``:
```
steinbro, line 33: This is not an unrealistic example, but it may be worth demonstrating that this technique is not just for custom scripts but can be used for key components of the image like the kernel itself.

I'm specifically thinking about illustrating why this product feedback request is unnecessary: https://warthogs.atlassian.net/browse/PF-6731

steinbro (commented):
I like this. I'd consider including a slightly more advanced/real-world example per my comment.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
+.. meta::
+    :description: How to include custom files in your image.
+
+.. _include-custom-files:
+
+Include custom files
```
jahn-junior: What do you think about renaming this to "Include local files"? I ask because, in my mind, a file can be custom whether you're sourcing it locally or remotely.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+====================
+
+Images may need to include arbitrary files, such as custom scripts or packages beyond
+what's available in package archives. You can include custom files in your image by
+placing them in your project directory and copying them to the overlay file system with
+a part that uses the :ref:`dump plugin <craft_parts_dump_plugin>`.
```
jahn-junior:
```suggestion
When crafting your image, you may need to copy custom scripts, pre-built packages,
or other static files from your local machine to the image's file system. This is done
by placing them in your image project directory and copying them to the overlay file
system with the :ref:`craft_parts_dump_plugin`.
```

> Images may need to include arbitrary files

* The user is the one including the file. The image contains it.
* 'arbitrary' commonly means that the files aren't necessary, which I imagine won't be common for users that made it to this page.

> You can include

Probably too much passive voice for an introductory paragraph.

> with a part that uses the dump plugin

I'm not entirely sure of this one, but I think we can leave out the explicit mention of parts. My thought is that if a user knows about parts and plugins, this will be implicit, and if they don't, then they'll be confused by our mention of parts anyways.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+
+To apply a configuration at first boot rather than baking it into the image, see
+:ref:`configure-instances-with-cloud-init`.
```
jahn-junior:
```suggestion
```

I think these guides are ultimately serving different needs, so I'd drop this link. It's also a little misleading since the linked guide "bakes in" a default cloud-init configuration similar to what's done here.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+-----------------
+
+In your project directory, create a directory to hold the custom files:
+
+.. code-block:: bash
+
+    mkdir my-files
+
+Place the files you want to include in this directory, preserving any directory structure
+that they should have in the final image.
```
jahn-junior: Some nits
```suggestion
Place the files you want to include in this directory, preserving the directory structure
they should have in the final image.
```

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+
+
+Copy the files to the image
+---------------------------
+
+Copy the files to the image with a new part that uses the :ref:`dump plugin
+<craft_parts_dump_plugin>` and the :ref:`organize <PartSpec.organize_files>` key. The
+part should run after the root file system is in place.
```
jahn-junior:
```suggestion
Copy the files to the image with a new part that uses the :ref:`craft_parts_dump_plugin` and the :ref:`organize <PartSpec.organize_files>` key. This part must be processed after the root file system is in place.
```

> the [dump plugin]()

I think this link text is redundant aside from a difference in capitalization, which is fine here.

> the part should run after

I took the language for my suggestion from the [after key's reference text](https://documentation.ubuntu.com/craft-parts/latest/common/craft-parts/reference/part_properties/#after).

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+    parts:
+      my-files:
+        after: [rootfs]
+        plugin: dump
+        source: my-files/
+        organize:
+          my-script.sh: (overlay)/usr/local/bin/my-script.sh
+
+The ``(overlay)/`` prefix ensures the file is copied to the overlay file system and
+appears in the final image at the specified path.
```
jahn-junior:
```suggestion
The ``(overlay)/`` prefix in the destination file path causes the file to be copied
into the image's overlay file system, which is where the image's contents are
prepared.
```
Normally, I wouldn't like the more explanatory tone in my suggestion, but I think it helps shed a little more light on what the overlay file system _is_.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Apply custom operations
```
jahn-junior:
```suggestion

Perform additional file operations
```
This may be a matter of taste, but I don't think of operations as something to be "applied". That sounds too passive for the process we're describing here IMO.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,73 @@
...
+-----------------------
+
+To apply custom operations to the files, use the ``override-overlay`` key in a separate
+part to run commands in the overlay file system.
+
+For example, to set the file permissions on ``/usr/local/bin/my-script.sh`` in the final
```
jahn-junior: Nit
```suggestion
For example, to modify the file permissions on ``/usr/local/bin/my-script.sh`` in the final
```

jahn-junior (commented):
Thanks! I like the shape of this.

I left a mix of high-level suggestions and little wording tweaks for my initial pass. If you rework the example like @steinbro suggested, I'll have another look.

docs/how-to/include-custom-files.rst
lengau: Well the dump plugin does also work with a remote source.

docs/how-to/include-custom-files.rst
lengau: Another good couple of imagecraft-specific examples would be:

- Custom sources files, apt config/preferences, etc.
- Things like fstab

lengau (approved):
LGTM with the caveats of JJ's and Daniel's comments. Once those are addressed to their satisfaction I'm happy to merge!

docs/how-to/include-custom-files.rst
jahn-junior: I thought about that, but I raised this because the rest of the guide seems to assume we're working with local files. Not saying we can't accommodate remote sources too, but it'd take a little more thinking on the page structure.

lengau: Cool I'll defer to you on that

docs/how-to/include-custom-files.rst
smethnani: Fair enough! I added an example to install a custom kernel. Or do you think the simple example should be totally replaced?

docs/how-to/include-custom-files.rst
smethnani: @steinbro I know you specifically asked for this link but I agree with JJ that it seems a bit misplaced, especially since the cloud init guide also bakes in files. I removed the link but would you rather we re-add it as a related reference at the bottom of the page or something like that?

docs/how-to/include-custom-files.rst
smethnani: I don't feel strongly either way but the guide does assume local files atm. We can change it back to "custom" if we add an example of using the dump plugin with a remote source but how different do you think the page structure should be in that case?

docs/how-to/include-custom-files.rst
steinbro: I'm OK omitting it; I'm just concerned across our docs that the guides suggest solutions to general problems even when subtly different requirements would lead to entirely different instructions.  Another example is configuring package archives, which will look different depending on whether the user wants the archives to be used in the build environment, in the final image, or both.

docs/how-to/include-custom-files.rst
steinbro: I like the new example, and think keeping the simple example is fine too.

docs/how-to/include-custom-files.rst
```diff
@@ -0,0 +1,92 @@
...
+    parts:
+      my-files-perms:
+        after: [my-files]
+        plugin: nil
+        override-overlay: |
+          chmod +x /usr/local/bin/my-script.sh
+
+This technique is not limited to custom scripts and can be used for key components of
+the image. For example, to install a custom kernel from a set of ``.deb`` files, copy
+the packages into the overlay and install them:
```
jahn-junior:
```suggestion
the image. The following parts, for example, install a custom kernel from a set of
local Debian files:
```
Making this statement more declarative since it isn't really an instruction.

docs/how-to/include-custom-files.rst
jahn-junior: I also like how the examples play together. The simpler example lets users focus on and understand the machinery itself, and the second scales it up to something you might see in production.

docs/how-to/include-custom-files.rst
jahn-junior: It'd probably warrant a fork in the page. Something like:

- Local files
  - Prepare the files
  - Copy the files to the image
- Remote files
- Perform additional file operations

It's obviously not any more complex of a task, and it's going to feel very similar for users. I'd only want the page structure to show the difference so a user doesn't click on an 'Include custom files' guide looking for guidance on adding remote files, only to see the first section and write it off as not fitting their use case.

Again, no worries for now. I think the current scope is just fine.

jahn-junior (approved):
LGTM!

PR #394, opened 2026-08-10
test: add core pi boot spread test

smethnani:
This PR adds a pi variant for the core spread boot tests.

QEMU doesn't run the Pi firmware so we need to extract and pass the kernel, initrd and device tree. We also need to invoke the command twice, once for install mode and another for run mode. Snapd switches this in cmdline.txt but qemu doesnt not read cmdline.txt (the image works as is with a physical device)

---

### DIFF
--- a/tests/spread/boot/core/pi/imagecraft.yaml
+++ b/tests/spread/boot/core/pi/imagecraft.yaml
@@ -1,0 +1,40 @@
+name: test
+version: "0.1"
+summary: test
+description: test
+base: bare
+build-base: ubuntu@24.04
+
+platforms:
+  arm64:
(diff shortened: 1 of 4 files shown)


lengau (approved):
Thanks! Can confirm this boots correctly on a raspberry pi

steinbro (approved):
I'm happy with this if @lengau has confirmed it works. I think the nested conditionals in tests/spread/boot/core/task.yaml suggest this file may be cleaner broken into separate tests, but that's not a blocker.

lengau:
@steinbro I agree about that spread test, but for the de-loopification work I'll be reorganizing those tests anyway (due to the need to cover some other cases).

PR #411, opened 2026-08-13
feat(pack): add partition size and by-number lookup helpers to gptutil

lengau:
Part 2 of a stack that makes GRUB installation work without loop devices.

To write into a partition of a disk image without mounting it, we need to know
where that partition starts and how big it is. `gptutil` could already list
partitions, but callers had to hand-roll the arithmetic to turn a partition
number into a byte offset and size.

This adds small helpers to do that lookup, plus unit tests. There are no callers
yet -- the code that uses them arrives later in the stack.

Also backfills tests for the existing helpers, which were only covered
indirectly by slow end-to-end tests.

IMAGECRAFT-176

### DIFF
--- a/imagecraft/pack/gptutil.py
+++ b/imagecraft/pack/gptutil.py
@@ -222,7 +222,7 @@ def _partition_node_number(
     """
     device = str(table.get("device", ""))
     node = str(partition.get("node", ""))
-    suffix = node[len(device) :] if node.startswith(device) else node
+    suffix = node.removeprefix(device)
     match = re.fullmatch(r"p?(\d+)", suffix)
     return int(match.group(1)) if match else None
 
--- a/imagecraft/pack/gptutil.py
+++ b/imagecraft/pack/gptutil.py
@@ -15,6 +15,7 @@
 """Utility functions for GPT-formatted disks."""
 
 import json
+import re
 import subprocess
 from pathlib import Path
 from typing import Any, cast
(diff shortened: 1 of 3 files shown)


imagecraft/pack/gptutil.py
```diff
@@ -202,12 +203,45 @@ def _get_partition_info(imagepath: Path, partname: str) -> dict[str, Any]:
...
     raise CraftError(f"No partition named {partname} in {imagepath}")
 
 
+def _partition_node_number(
+    table: dict[str, Any], partition: dict[str, Any]
+) -> int | None:
+    """Return the partition number sfdisk reported for a partition entry.
+
+    sfdisk only lists partitions that are in use, so a table with gaps in its
+    numbering (possible for GPT, which lets users pick partition numbers) has
```
mwhudson, line 218: Pretty sure it's possible for MBR primary partitions too?

tests/unit/pack/test_gptutil.py
```diff
@@ -208,6 +208,168 @@ def test_get_partition_sector_offset(mocker, tmp_path):
     assert gptutil.get_partition_sector_offset(tmp_path, "rootfs") == 526336
 
 
+_TWO_PARTITION_SFDISK_JSON = """
```
mwhudson, line 211: is it possible to write a slightly less fake test that makes a small disk image, gives it some partitions and then runs sfdisk -J on that? I think that should still be pretty fast to run.

mwhudson (commented):
LGTM, this stuff is so fiddly and annoying. A couple of questions.

imagecraft/pack/gptutil.py
```diff
@@ -202,12 +203,45 @@ def _get_partition_info(imagepath: Path, partname: str) -> dict[str, Any]:
...
+    entries whose number doesn't match their position in the list. The number
+    has to be recovered from the node name, which is the device path with the
+    partition number (optionally prefixed by "p") appended.
+    """
+    device = str(table.get("device", ""))
+    node = str(partition.get("node", ""))
+    suffix = node[len(device) :] if node.startswith(device) else node
```
bepri:
```suggestion
    suffix = node.removeprefix(device)
```

bepri (approved):

steinbro (approved):

imagecraft/pack/gptutil.py
lengau: In theory yes. In practice we don't (currently at least) support `partition-number` for MBR so you can't skip a partition number.

tests/unit/pack/test_gptutil.py
lengau: df483ae8dcf942ff34cac15f7b3e6e2082d20b6a

lengau:
Bypassing unrelated snap pack issue.

PR #412, opened 2026-08-13
fix(pack): mark system-data bootable when there is no system-boot

lengau:
Part 3 of a stack that makes GRUB installation work without loop devices.

A BIOS/MBR image needs a partition flagged bootable for the firmware to hand
over to it. We previously only ever set that flag on the `system-boot`
partition, so an MBR image with no separate boot partition -- everything on the
rootfs, which is a perfectly normal layout -- ended up with no bootable
partition at all and would not boot.

Fall back to marking the `system-data` partition bootable when there is no
`system-boot` partition.

IMAGECRAFT-176

### DIFF
--- a/tests/unit/pack/test_mbrutil.py
+++ b/tests/unit/pack/test_mbrutil.py
@@ -206,10 +206,6 @@ def test_create_mbr_layout_calls_sfdisk(mocker, tmp_path, volume):
 
 
 def test_create_mbr_layout_no_system_boot_marks_system_data_bootable(mocker, tmp_path):
-    # Regression test: layouts with no system-boot partition (e.g. grub-pc
-    # BIOS-only images where core.img lives in the MBR gap) must still mark
-    # a partition bootable, or SeaBIOS has no active partition to boot from
-    # and hangs at "Booting from Hard Disk...".
     layout = MBRVolume.unmarshal(_VOLUME_SINGLE)
     mocked_run = mocker.patch("imagecraft.pack.mbrutil.subprocess.run", autospec=True)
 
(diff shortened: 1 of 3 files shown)


tests/unit/pack/test_mbrutil.py
```diff
@@ -205,6 +205,87 @@ def test_create_mbr_layout_calls_sfdisk(mocker, tmp_path, volume):
...
+
+    stdin = mocked_run.call_args.kwargs["input"]
+    partition_lines = [line for line in stdin.splitlines() if line.startswith("start=")]
+    bootable_lines = [line for line in partition_lines if "bootable" in line]
+
+    assert len(partition_lines) == 3
+    assert len(bootable_lines) == 1
+    # It must be the first partition, which is the one holding the root
+    # filesystem GRUB was pointed at.
+    assert bootable_lines[0] == partition_lines[0]
```
bepri, line 268: I don't understand how this is making sure that any particular partition is preferred. Does it just depend on the ordering of the entries in your mock input? Could we make the comparison more explicit somehow?

tests/unit/pack/test_mbrutil.py
```diff
@@ -205,6 +205,87 @@ def test_create_mbr_layout_calls_sfdisk(mocker, tmp_path, volume):
...
+        sector_size=512,
+        layout=layout,
+    )
+
+    stdin = mocked_run.call_args.kwargs["input"]
+    partition_lines = [line for line in stdin.splitlines() if line.startswith("start=")]
+    bootable_lines = [line for line in partition_lines if "bootable" in line]
+
+    assert len(bootable_lines) == 1
+    assert bootable_lines[0] == partition_lines[0]
```
bepri, line 286: Same question here

tests/unit/pack/test_mbrutil.py
lengau: No actually that's a really good question. I remember explicitly making this decision but now looking back I don't remember why. `system-boot` should be required in order to install grub.

PR #414, opened 2026-08-13
feat(pack): add imgfs for unprivileged partition image manipulation

lengau:
Part 4 of a stack that makes GRUB installation work without root privileges.

Adds `imagecraft.pack.imgfs`, a small wrapper for reading and writing files
inside a partition of a disk image *in place*, without mounting it, attaching a
loop device or being root. ext2/3/4 goes through `debugfs` and FAT through
`mtools`, both of which understand their filesystems well enough to operate on a
plain file at an offset.

There are no callers yet -- the GRUB code that uses this arrives later in the
stack. This lands separately so the new module can be reviewed and tested on its
own.

### A note on `debugfs`

Writing the tests turned up a trap worth flagging: **`debugfs` exits 0 even when
it could not open the filesystem at all.** Pointing it at a corrupt image prints

```
Attempt to read block from filesystem resulted in short read while trying to open ...
/tmp/x.img: Filesystem not open
```

and then returns success. So the obvious `check=True` / `CalledProcessError`
handling never fires, and every write silently does nothing. `debugfs_run`
therefore also scans the output for those markers. The tests cover this
explicitly.

### DIFF
--- a/imagecraft/pack/imgfs.py
+++ b/imagecraft/pack/imgfs.py
@@ -1,0 +1,492 @@
+# Copyright 2025 Canonical Ltd.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License version 3 as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Low-privilege partition image manipulation.
+
+These helpers let callers read and write files inside FAT and ext2/3/4
+partitions *embedded in a raw disk image* without mounting anything,
+attaching loop devices, chrooting, or spinning up a VM. This lets imagecraft
+run in unprivileged containers that can't do any of those things.
+
+FAT partitions are manipulated in place with ``mtools`` using its
+``image@@offset`` addressing (no extraction needed). Ext partitions are
+manipulated with ``debugfs``, which only understands whole filesystem
+images, so the partition is first extracted to a temporary file with
+``dd``, edited, then written back with ``dd``.
+"""
+
+import contextlib
+import os
+import posixpath
+import re
+import shutil
+import subprocess
+import tempfile
+from collections.abc import Iterator
+from pathlib import Path
+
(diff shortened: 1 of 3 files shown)


imagecraft/pack/imgfs.py
```diff
@@ -0,0 +1,492 @@
...
+    and a path containing a newline injects an entirely new request.
+    """
+    if _DEBUGFS_UNSAFE_PATH_RE.search(path):
+        raise errors.GRUBInstallError(
+            f"Refusing to pass unsafe path to debugfs: {path!r}"
+        )
+    return f'"{path}"'
+
+
+def debugfs_run(image: Path, commands: list[str]) -> str:
```
mwhudson, line 173: This feels like it should be internal API somehow

imagecraft/pack/imgfs.py
```diff
@@ -0,0 +1,492 @@
...
+        raise errors.GRUBInstallError(
+            f"{src} is too small: needs {count * bs} bytes at offset {skip * bs}"
+        )
+    _dd_copy(src=src, dst=dst, bs=bs, skip=skip, seek=seek, count=count)
+
+
+@contextlib.contextmanager
+def edit_ext_partition(
+    imagepath: Path, offset_sectors: int, size_sectors: int
+) -> Iterator[Path]:
```
mwhudson, line 126: Maybe I have too much OOP brain but did you consider having this contextmanger yield some kind of DebugFS object that would have read/write/exists/... methods?

mwhudson (commented):
Here are some slightly incoherent comments!

It's a bit hard to review this without knowing the scope of usage. Looking at the later branches it seems that debugfs_mkdir_p, mmd_p and (especially) debugfs_run could be private API.

I also get confused about the dependencies between the various steps of this. Why does grub installation have to run after the image is created? If it's just to have access to the filesystem uuid, that can be preallocated surely? I guess for BIOS we have to write to the MBR... 

The use of e.g. `debugfs_read_file` to read grub out of the image is probably extremely correct but the files are present in the prime directory too and we wouldn't have to use debugfs nearly as much if we just read them from there. (In general, dd-ing the rootfs out of the image, using debugfs on it, then dd-ing it back in seems a bit inefficient?).

(Do you have a design document you are working off for all of this?)

imagecraft/pack/imgfs.py
lengau: I'm rethinking this. I don't really want to have custom stuff for each filesystem either, so I'm looking at making a wrapper for fusepy that'll let me do this. Then the limiting factor will be having a fuse FS.

lengau:
Superseded by #428

PR #428, opened 2026-08-24
feat(utils): add FUSE partition and volume mounting utilities

lengau:
## Description

Adds FUSE-based partition and volume mounting utilities under `imagecraft.utils.mount` using `fuse2fs`, `fusefat`, and `fusefile`, enabling mounting of ext4, FAT, and composite GPT/MBR volume hierarchies without kernel loop devices.

Also fixes a race condition where `/dev/loopXpY` partition device nodes are accessed before udev finishes creating them after `losetup --partscan`.

### Key Changes
- **FUSE Mounts (`imagecraft.utils.mount`)**: `mount_partition()` and `mount_volume()` context managers supporting ext4 (`fuse2fs`), FAT (`fusefat`), and partition offsets via `fusefile`.
- **Loop Device Partition Sync (`imagecraft.services.image`)**: Polls for `/dev/loopXpY` nodes after `losetup --partscan`.
- **Packaging & Tests**: Stages FUSE utilities in snapcraft and adds unit and integration tests.

IMAGECRAFT-176

---

### DIFF
--- a/.github/workflows/qa.yaml
+++ b/.github/workflows/qa.yaml
@@ -47,9 +47,9 @@ jobs:
     needs: filter
     with:
       source-files: ${{ needs.filter.outputs.source-files }}
-      fast-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm"]'
+      fast-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm", "ubuntu-26.04-arm"]'
       fast-test-python-versions: '["3.12"]'
-      slow-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm"]'
+      slow-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm", "ubuntu-26.04-arm"]'
       slow-test-python-versions: '["3.12"]'
       lowest-python-version: ""
       pytest-markers: requires_root
--- a/imagecraft/utils/mount.py
+++ b/imagecraft/utils/mount.py
@@ -575,6 +575,13 @@ class CompositeMount(BaseMount):
             )
 
 
+_DEFAULT_ROLE_MOUNTS: dict[Role, str] = {
+    Role.SYSTEM_DATA: "",
+    Role.SYSTEM_BOOT: "boot/efi",
+    Role.SYSTEM_SEED: "var/lib/snapd/seed",
+}
+
+
 def mount_volume(
     volume: Volume,
     disk_path: Path,
@@ -610,18 +617,10 @@ def mount_volume(
         if not filesystem:
             continue
 
-        # Determine relative mount path
-        rel_path: str
-        if item.name in mountpoint_overrides:
-            rel_path = mountpoint_overrides[item.name]
-        elif item.role == Role.SYSTEM_DATA:
-            rel_path = ""
-        elif item.role == Role.SYSTEM_BOOT:
-            rel_path = "boot/efi"
-        elif item.role == Role.SYSTEM_SEED:
-            rel_path = "var/lib/snapd/seed"
-        else:
-            rel_path = f"mnt/{item.name}"
+        rel_path = mountpoint_overrides.get(
+            item.name,
+            _DEFAULT_ROLE_MOUNTS.get(item.role, f"mnt/{item.name}"),
+        )
 
         # Calculate byte offset and size
         if is_gpt:
--- a/imagecraft/utils/mount.py
+++ b/imagecraft/utils/mount.py
@@ -475,10 +475,11 @@ def mount_partition(
     :param fakeroot: If True, pass fakeroot option to FUSE (ext only).
     :raises errors.MountError: If the filesystem type is unsupported.
     """
-    if isinstance(filesystem, FileSystem):
-        fs_str = filesystem.value.lower()
-    else:
-        fs_str = str(filesystem).lower()
+    fs_str = (
+        filesystem.value.lower()
+        if isinstance(filesystem, FileSystem)
+        else str(filesystem).lower()
+    )
     if fs_str in ("ext2", "ext3", "ext4"):
         return ExtFuseMount(
             imagepath,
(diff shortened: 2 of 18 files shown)


tests/integration/utils/test_mount.py
```diff
@@ -0,0 +1,234 @@
...
+    "fusermount3": "fuse3",
+}
+
+
+@pytest.fixture(autouse=True, scope="module")
+def check_required_executables() -> None:
+    missing: list[str] = [
+        f"{binary} (install package: {pkg})"
+        for binary, pkg in REQUIRED_EXECUTABLES.items()
+        if shutil.which(binary) is None
```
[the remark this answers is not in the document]
lengau: If it's not on the user's path we don't want it to be used.

lengau:
Test failure is unrelated (loop device that will be pulled out with IMAGECRAFT-192)

.github/workflows/qa.yaml
```diff
@@ -47,9 +47,9 @@ jobs:
     needs: filter
     with:
       source-files: ${{ needs.filter.outputs.source-files }}
-      fast-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm"]'
+      fast-test-platforms: '["ubuntu-24.04", "ubuntu-24.04-arm", "ubuntu-26.04-arm"]'
```
lengau, line 50: Added this because Ubuntu 24.04 on ARM doesn't have fusefat

imagecraft/utils/mount.py
```diff
@@ -0,0 +1,625 @@
...
+                last_err = err
+            else:
+                return
+        time.sleep(0.1)
+
+    if last_err is not None:
+        raise last_err
+
+
+def _fuse_command(command: Sequence[str], description: str) -> None:
```
bepri:
```suggestion
def _try_fuse_command(command: Sequence[str], *, err_msg: str) -> None:
```

I think this signature will make it a little bit more obvious what it's doing at callsites

imagecraft/utils/mount.py
```diff
@@ -0,0 +1,625 @@
...
+    GPTVolume,
+    PartitionSchema,
+    Role,
+    Volume,
+)
+from imagecraft.pack import gptutil
+from imagecraft.subprocesses import run
+
+
+def _unmount_path(
```
bepri, line 54: I think this function ought to use the `_fuse_command` wrapper too, there's at least one other point (line 310) where you're already more or less doing the same thing

tests/integration/utils/test_mount.py
```diff
@@ -0,0 +1,320 @@
...
+        os_release = platform.freedesktop_os_release()
+        return os_release.get("VERSION_CODENAME") == "noble"
+    except (AttributeError, OSError):
+        pass
+    os_release_path = Path("/etc/os-release")
+    if os_release_path.exists():
+        for line in os_release_path.read_text().splitlines():
+            if line.startswith("VERSION_CODENAME="):
+                return line.split("=", 1)[1].strip("\"'") == "noble"
+    return False
```
bepri: Why is this check so much more rigorous than the others? Can it be replaced by my suggested fixture too?

tests/integration/utils/test_mount.py
```diff
@@ -0,0 +1,320 @@
...
+            dict.fromkeys(
+                pkg
+                for binary, pkg in REQUIRED_EXECUTABLES.items()
+                if shutil.which(binary) is None
+            )
+        )
+        pytest.fail(
+            f"Missing required executables for FUSE integration tests:\n  - {missing_str}\n"
+            f"Install them with:\n  sudo apt-get install {missing_pkgs}"
+        )
```
bepri: Can't we just include these in the makefile and get rid of this fixture entirely?

tests/integration/utils/test_mount.py
```diff
@@ -0,0 +1,320 @@
...
+
+        # Write to EFI / boot partition (vfat, nested under /boot/efi)
+        efi_root = rootfs / "boot" / "efi"
+        if (efi_root / "EFI").is_file() and (efi_root / "EFI").stat().st_size == 0:
+            (efi_root / "EFI").unlink()
+
+        efi_boot = efi_root / "EFI" / "BOOT"
+        efi_boot.mkdir(parents=True, exist_ok=True)
+        (efi_boot / "grubx64.efi").write_bytes(b"GRUB_BINARY")
```
steinbro, line 278: This is unusually expressive and elegant for Python code.

tests/integration/utils/test_mount.py
lengau: Lol because I had Gemini do this. IDK what suggested fixture you're talking about though?

imagecraft/utils/mount.py
lengau: Good point. IDK why I didn't follow my own rule of "start a function/method name with a verb."

imagecraft/utils/mount.py
lengau: It doesn't really save us much, but I've done so.

tests/integration/utils/test_mount.py
bepri: Oh I originally was going to propose that the "are we on Noble?" helper be turned into a fixture instead since you did it similarly in multiple places, but then I decided that it wasn't super important given how few times it's needed. I forgot to update this comment hehe

```py
@pytest.fixture
def noble_only() -> None:
  if distro.codename() != "noble":
    pytest.skip("Test must be run on Noble.")
```

lengau: Well in this case we need the opposite :-)

tests/integration/utils/test_mount.py
lengau: We do include them in the Makefile, but we need this anyway because some of the packages don't exist everywhere...

bepri (approved):

steinbro (approved):
I'm OK with these changes. As discussed, a consequence will be more work required to add support for additional filesystems that happen to have kernel support, but (a) we don't expect to be adding new filesystem types frequently and (b) resorting to loopback devices on non-LXD providers will still be a fallback option.

